PR #30, opened 2023-11-14
Adds Object Segmentation Tool

marklundin:
This PR adds a simple object segmentation mask tool, to filter out foreground and background components of the 3DGS. it works in a similar 2d space as the brush tool, but is useful as a first pass, which can be refined using the brush tool. 

Because it's using NN it's output is as good as the underlying model, which has varying degrees of success. Seems to work better on people, than other arbitrary ill-defined object.

See video demo
https://www.loom.com/share/b996905904184c2884386770a12ff04c?sid=2389555b-8bb5-49cf-8313-9018e1303d93

image


package.json
```diff
@@ -72,5 +72,9 @@
         "rollup-plugin-visualizer": "^5.9.2",
         "serve": "^14.2.0",
         "tslib": "^2.5.3"
+    },
+    "dependencies": {
```
slimbuck, line 76: are these `devDependencies`?

src/scene-config.ts
```diff
@@ -50,7 +50,7 @@ const sceneConfig = {
         dampingFactor: 0.2,
         minPolarAngle: 0,
         maxPolarAngle: 2.8,
-        minZoom: 0.01,
+        minZoom: 0.0001,
```
slimbuck, line 53: think this might be too small tbh. once you get to this level of zoom your camera becomes immobile. we need better controls.

src/ui/control-panel.ts
```diff
@@ -1,5 +1,13 @@
 import { EventHandler } from 'playcanvas';
 import { BooleanInput, Button, Container, Label, NumericInput, Panel, RadioButton, SelectInput, SliderInput, VectorInput } from 'pcui';
+import { version as supersplatVersion } from '../../package.json';
+import logo from './playcanvas-logo.png';
```
slimbuck, line 4: these two were removed in a previous PR

package.json
marklundin: nope, these are part of the source

src/scene-config.ts
marklundin: Ok yep, this was unintentional, and only used in my testing

package.json
slimbuck: oh interesting. what do they provide at runtime?

marklundin: So these do the object segmentation at runtime. I generally keep `devDependancies` as build tools etc, and anything used in the code itself in dependancies, but let me know what works best

willeastcott: If the app builds+runs with both of these in devDependencies, they should live there.

willeastcott:
This was a super-cool experiment, but I don't think we'll take this approach for segmentation. Closing for now.

johnbanq:
> This was a super-cool experiment, but I don't think we'll take this approach for segmentation. Closing for now.

@willeastcott Hi! Can I get more details on why this approach was not considered?

willeastcott:
I can't really remember. Maybe because it didn't isolate an object but just created a mask which would also select gaussians behind the segmented object? Or maybe the segmentation wasn't reliable/robust enough at the time? Anyway, it didn't seem ready for primetime, so I closed it.

PR #38, opened 2023-11-15
Add scale limit

slimbuck:
Limit scale input to positive values.

### DIFF
--- a/src/ui/control-panel.ts
+++ b/src/ui/control-panel.ts
@@ -610,7 +610,9 @@ class ControlPanel extends Container {
         const scaleInput = new NumericInput({
             class: 'control-element-expand',
             precision: 2,
-            value: 1
+            value: 1,
+            min: 0.01,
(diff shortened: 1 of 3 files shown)


src/ui/control-panel.ts
```diff
@@ -610,7 +610,9 @@ class ControlPanel extends Container {
         const scaleInput = new NumericInput({
             class: 'control-element-expand',
             precision: 2,
-            value: 1
+            value: 1,
+            min: 0.01,
```
mvaligursky, line 614: if they can scale up by 10k, why not down as well?

mvaligursky (approved):

src/ui/control-panel.ts
slimbuck: I tried, but `min` must be a multiple of `precision` (which itself must be 2 for natural scaling).

PR #52, opened 2023-12-13
Avoid dividing by zero

kemchenj:


slimbuck:
Thanks @kemchenj, this is great!

Another way we might solve the divide by zero is to add a helper function:
```javascript
        const normalize = (x: number, min: number, max: number) => {
            return (max - min < 0.0001) ? 0 : (x - min) / (max - min);
        };
```

And then modify the calls to use it:
```javascript
            this.position[i] = pack111011(
                normalize(x[i], px.min, px.max),
                normalize(y[i], py.min, py.max),
                normalize(z[i], pz.min, pz.max)
            );
```

Because if max === min we probably don't want to perform a divide at all!

src/splat-convert.ts
```diff
@@ -278,17 +282,17 @@ class Chunk {
         // pack
         for (let i = 0; i < this.size; ++i) {
             this.position[i] = pack111011(
-                (x[i] - px.min) / (px.max - px.min),
-                (y[i] - py.min) / (py.max - py.min),
-                (z[i] - pz.min) / (pz.max - pz.min)
+                normalize(x[i], px.max, px.min),
+                normalize(y[i], py.max, py.min),
+                normalize(z[i], pz.max, pz.min)
```
slimbuck, line 287:
```suggestion
                normalize(x[i], px.min, px.max),
                normalize(y[i], py.min, py.max),
                normalize(z[i], pz.min, pz.max)
```

src/splat-convert.ts
```diff
@@ -278,17 +282,17 @@ class Chunk {
...
 
             this.rotation[i] = packRot(rot_0[i], rot_1[i], rot_2[i], rot_3[i]);
 
             this.scale[i] = pack111011(
-                (scale_0[i] - sx.min) / (sx.max - sx.min),
-                (scale_1[i] - sy.min) / (sy.max - sy.min),
-                (scale_2[i] - sz.min) / (sz.max - sz.min)
+                normalize(scale_0[i], sx.max, sx.min),
+                normalize(scale_1[i], sy.max, sy.min),
+                normalize(scale_2[i], sz.max, sz.min)
```
slimbuck, line 295:
```suggestion
                normalize(scale_0[i], sx.min, sx.max),
                normalize(scale_1[i], sy.min, sy.max),
                normalize(scale_2[i], sz.min, sz.max)
```

slimbuck (approved):
Thanks!!

PR #64, opened 2024-01-30
Add transform gizmos

slimbuck:
Screenshot 2024-01-30 at 15 31 38

This PR:
- updates super-splat to use engine [v1.68.0](https://github.com/playcanvas/engine/releases/tag/v1.68.0)
- adds transform gizmos for orientating the splat in the scene
    - integrates with undo/redo
    - local/world space toggle
    - using the new 🔥 https://github.com/playcanvas/engine/pull/5928
- adds a home grid visual
- a few UI tweaks
- removes legacy unused bits from super-splat, like draco decoder


marklundin:
Scale gizmo sometimes gets a bit lost in the background. Not sure if we can make these stand out a bit more 

image

willeastcott:
> Scale gizmo sometimes gets a bit lost in the background. Not sure if we can make these stand out a bit more
Ha, I literally just pointed that out to @slimbuck. I think the Editor always renders its gizmos over (replaces) scene pixels - that doesn't seem to be happening here...

kpal81xd:
> > Scale gizmo sometimes gets a bit lost in the background. Not sure if we can make these stand out a bit more
> 
> Ha, I literally just pointed that out to @slimbuck. I think the Editor always renders its gizmos over (replaces) scene pixels - that doesn't seem to be happening here...

The gizmo is rendered in its own layer so might be some layer ordering issues

slimbuck:
I just checked and gizmos are rendering last. I think we need to increase semitransparency for active and inactive states. Could this be exposed as two values?

src/events.ts
```diff
@@ -0,0 +1,30 @@
+import { EventHandler } from "playcanvas";
```
willeastcott, line 1: Any reason not to just import Events from `@playcanvas/observer`?

src/tools/rotate-tool.ts
```diff
@@ -0,0 +1,19 @@
...
+import { EditHistory } from '../edit-history';
+import { Scene } from '../scene';
+
+class RotateTool extends TransformTool {
+    ToolName = 'Rotate';
+
+    constructor(events: Events, editHistory: EditHistory, scene: Scene) {
+        const gizmo = new RotateGizmo(scene.app, scene.camera.entity.camera, scene.gizmoLayer);
+
+        gizmo.coordSpace = 'local';
```
willeastcott, line 13: Is this line redundant?

src/ui/control-panel.ts
```diff
@@ -618,28 +533,34 @@ class ControlPanel extends Panel {
...
+                    events.fire('focusCamera');
                 } else if (e.key === 'B' || e.key === 'b') {
-                    toggle(brushSelection);
+                    events.fire('tool:activate', 'BrushSelection');
                 } else if (e.key === 'I' || e.key === 'i') {
-                    this.events.fire('invertSelection');
+                    events.fire('invertSelection');
                 } else if (e.key === '[') {
-                    brushSelection.smaller();
+                    events.fire('BrushSelection:smaller');
```
willeastcott, line 561: For some reason, I hate events with capital letters. Personal preference. 😄

src/ui/tooltip.ts
```diff
@@ -0,0 +1,50 @@
...
+class Tooltip extends Container {
+    target: Element;
+    align: string;
+    content: Label;
+
+    constructor(args: any = {}) {
+        args = Object.assign(args, {
+            class: 'tooltip',
+            hidden: true
+        });
```
willeastcott, line 12:
```suggestion
        args = {
            ...args,
            class: 'tooltip',
            hidden: true
        };
```

src/ui/tooltip.ts
```diff
@@ -0,0 +1,50 @@
...
+
+        super(args);
```
willeastcott, line 14:
```suggestion
        super({
            ...args,
            class: 'tooltip',
            hidden: true
        });
```

src/ui/tooltip.ts
```diff
@@ -0,0 +1,50 @@
...
+
+        this.target = args.target;
+        this.align = args.align || 'right';
```
willeastcott, line 17:
```suggestion
        this.align = args.align ?? 'right';
```

willeastcott (approved):
Super-duper PR - love it!

src/events.ts
slimbuck: Seems a pretty arbitrary choice. Is there a reason to prefer observer events?

src/tools/rotate-tool.ts
slimbuck: yes, left over testing code. thanks.

src/ui/control-panel.ts
slimbuck: There were reasons, but I don't love it either. Made it start small.

src/ui/tooltip.ts
slimbuck: Since args is usually referenced after invoking `super()`, I prefer to resolve args it as it is now.

src/events.ts
willeastcott: Well, mainly because I'm anticipating SuperSplat adopting our Observer, which subclasses the Observer Events class.

src/ui/tooltip.ts
slimbuck: Changed to it to use spread syntax

src/events.ts
slimbuck: Yeah makes sense, I'll make that change in a followup PR.

PR #68, opened 2024-02-15
Fix gizmo position after scene edit

slimbuck:
This PR:
- recalculates the gizmo position upon scene edit to keep it centered on the splat
- update npm dependencies, including latest engine
- bump app to v0.15.0


src/splat.ts
```diff
@@ -84,6 +78,31 @@ class Splat extends Element {
...
+
+        // update splat data local bound
+        const opacity = this.splatData.getProp('opacity');
+        this.splatData.calcAabb(localBound, (i: number) => {
+            return opacity[i] > -1000;
+        });
+
+        // calculate meshinstance aabb (transformed local bound)
+        const meshInstance = this.root.gsplat.instance.meshInstance;
+        meshInstance._aabb.setFromTransformedAabb(localBound, this.entity.getWorldTransform());
```
willeastcott, line 94: No way to avoid non-API calls?

willeastcott (approved):

src/splat.ts
slimbuck: I wasn't able to use public API. This needs addressing with @mvaligursky.

willeastcott: Is there an engine issue. Just wanna make sure we don't forget this.

slimbuck: Good point, I've created https://github.com/playcanvas/engine/issues/6055.

Issue #50, opened 2023-12-11
[Feature Request] Points that indicate approximate range


pjbaron:
Currently the points are all the same size and colour. It is frequently the case that we have folds or awkward shapes where it is very difficult to arrange the model to eliminate the possibility of points on 'the other side' overlapping our selection area.
This suggestion is to draw the points as small circles, with radii ranging from perhaps 6 pixels at the very largest to 1 pixel as the smallest, based on distance.
This would enable quick eyeball identification of problematic overlaps, without needing to jiggle the camera up and down to view the parralax motions.
I believe it would work very well in combination with the previously suggested 'depth' selection mode.

slimbuck:
We've added a "Rings" camera mode. Please check it out, it should be what you're looking for, but please let us know if not.

(This update will be released soon in v0.16.0).

Issue #47, opened 2023-12-11
[Feature Request] Click to select a Gaussian's point individually


pjbaron:
After some thorough cleaning, I tend to find that there are a number of large gaussian splats sticking out from surfaces and edges where an origin point is embedded somewhere in a tight group of valid points.
This suggestion is that it be possible to switch to a 'gaussian select' mode, and then individual mouse clicks will highlight the gaussian and it's origin point.
The approach would need to take into account alpha (ignore all overlapping gaussians which are too faint) and the gaussian value would need to be sufficiently strong (80%+) to avoid too many overlaps, but weak enough so that it's not necessary to find the middle too accurately (which would defeat the purpose).
Given that often it's just a single big splat sticking out, I'd be happy if it only ever picked the gaussian which is contributing the highest amount of colour to the region clicked... combining both requirements into one.

Three large 'bad' splats with difficult to isolate points:
![image](https://github.com/playcanvas/super-splat/assets/614817/7eb345f8-d0b8-4120-9168-ecc40b7a91e5)

Issue #29, opened 2023-11-13
[Feature request] Change selection depth for easier cleaning


TimeScience:
This tool is really awesome. Thanks so much for making it.

However, I'm finding it super hard to easily clean up models because the selection tool selects everything under the selection tool, without any reference to depth; and since the gaussian models include background in all directions, it is quite hard to only select the noise points I want to delete that are near the camera, without also selecting everything behind it that I don't want to get rid of. If it was possible to set a depth of selection (maybe as in relation to the distance from the camera?) this might make cleaning up the model way easier. 

I suspect the really easiest way to make cleanup easy would be to make a webVR interface so users could just go into the model in 3D and mark up and remove the bad splats by hand, but I'm assuming that is not technically easy to implement :)

slimbuck:
Hi @TimeScience ,

That's not a bad idea at all and actually in principle not too difficult to add either. What do you think @willeastcott ?

Thanks!

TimeScience:
Have a look at this file as a good example... there's a bunch of random sky garbage above the chess set, but if you try and remove 
 it, it is quite hard to get the camera positioned right in super-splat viewer. If once could set the "select distance" with a slider of some sort, I imagine this would be much easier.

[https://filesender.aarnet.edu.au/?s=download&token=84b3eddf-7138-4254-85a0-23ee43a38de4](https://filesender.aarnet.edu.au/?s=download&token=84b3eddf-7138-4254-85a0-23ee43a38de4) (link expires 23-Dec 2023)

willeastcott:
Sounds like a useful feature to me.

willeastcott:
Maybe a good way to do this is not with depth explicitly, but by editing the clip planes in SuperSplat?

TimeScience:
If it was easier to let the user draw a cubical or spherical bounding box, that might be a quite good solution. Essentially you want a way for people to be able to rapidly select & delete the random splats that GS drops all over the place. So if there was a box or sphere they could drag around in 3D space and just delete the contents, that might work well.

marklundin:
> If it was easier to let the user draw a cubical or spherical bounding box, that might be a quite good solution. Essentially you want a way for people to be able to rapidly select & delete the random splats that GS drops all over the place. So if there was a box or sphere they could drag around in 3D space and just delete the contents, that might work well.

There is actually a sphere selection tool you can use to filter things out. It's controlled via the params on the side, so unfortunately no gizmo at the moment

[...]

PR #83, opened 2024-04-19
Add new splat mode, add hiding support

slimbuck:
Fixes: #50, #29, #47

This PR:
* introduces a new rendering/selection mode called "Rings", which allows selection by splat extents instead of splat centers
* adds an option to hide selected splats and unhide them
* adds a new picker selection tool to extend rect and brush
* uses a state array to track each splat state, instead of updating splat opacity
* double click the scene to center the orbit camera around the target point
* update npm dependencies, including engine [v1.70.0](https://github.com/playcanvas/engine/releases/tag/v1.70.0)

Screenshot 2024-04-19 at 12 01 12


src/camera.ts
```diff
@@ -259,17 +293,29 @@ class Camera extends Element {
 
         // in with the new
         const pixelFormat = PIXELFORMAT_RGBA8;
+        const samples = this.scene.config.camera.multisample ? device.maxSamples : 1;
```
mvaligursky, line 296: any benefits to use multisampling in this tool?

src/camera.ts
```diff
@@ -259,17 +293,29 @@ class Camera extends Element {
...
         this.entity.camera.renderTarget = renderTarget;
         this.entity.camera.camera.horizontalFov = width < height;
+
+        // create pick mode render target
+        this.pickModeColorBuffer = createTexture(width, height, pixelFormat);
+        this.pickModeRenderTarget = new RenderTarget({
+            colorBuffer: this.pickModeColorBuffer,
+            depth: false,
+            flipY: false,
+            samples: samples,
```
mvaligursky, line 316: I assume the picker rendering here renders some IDs. Is it compatible with multisampling? I assume the resolve would average IDs.

src/editor.ts
```diff
@@ -0,0 +1,675 @@
...
+            const state = splatData.getProp('state') as Uint8Array;
+            const x = splatData.getProp('x');
+            const y = splatData.getProp('y');
+            const z = splatData.getProp('z');
+
+            vec.set(axis[0], axis[1], axis[2]);
+            vec2.set(axis[0] * distance, axis[1] * distance, axis[2] * distance);
+
+            // transform the plane to local space
+            mat.invert(splatDef.element.worldTransform);
```
mvaligursky, line 401: Not entirely sure what matrix gets inverted here .. just making sure this needs to take place in the loop

mvaligursky (approved):
Awesome!

src/camera.ts
slimbuck: nope, it's disabled in config https://github.com/playcanvas/super-splat/blob/main/src/scene-config.ts#L23, but I decided to keep support in the code for now.

src/camera.ts
slimbuck: probably not compatible with resolve, but as mentioned above it's disabled. i might completely remove multisampling support in the next update.

src/editor.ts
slimbuck: The loop here is over instances of GSplat in the scene, not individual splats. The loop over splats happens in `processSelection` below. 👍🏻

src/splat.ts
```diff
@@ -81,13 +204,46 @@ class Splat extends Element {
...
-        const opacity = this.splatData.getProp('opacity');
-        this.splatData.calcAabb(localBound, (i: number) => {
-            return opacity[i] > -1000;
-        });
+        localBound.center.set((minx + maxx) * 0.5, (miny + maxy) * 0.5, (minz + maxz) * 0.5);
+        localBound.halfExtents.set((maxx - minx) * 0.5, (maxy - miny) * 0.5, (maxz - minz) * 0.5);
+
+        // this.splatData.calcAabb(localBound, (i: number) => {
+        //     return opacity[i] > -1000;
+        // });
```
willeastcott, line 246: Delete?

src/tools/picker-selection.ts
```diff
@@ -0,0 +1,37 @@
...
+
+class PickerSelection {
+    events: Events;
+    root: HTMLElement;
+
+    constructor(events: Events, parent: HTMLElement) {
+        this.root = document.createElement('div');
+        this.root.id = 'select-root';
+
+        this.root.onmousedown = (e: MouseEvent) => {
```
willeastcott, line 11: Can we aim to stick to pointer events in our apps from now on?

willeastcott (approved):
Can't wait for this to be released.

src/tools/picker-selection.ts
slimbuck: Yes! I will update everything to pointer events in a separate PR.

PR #86, opened 2024-04-24
Disable contextmenu when picker-selection activated

gotoeasy:
Disable contextmenu when picker-selection activated


slimbuck (approved):
Lovely, thank you! <3

PR #92, opened 2024-05-18
Associate PLY file with the SuperSplat PWA

### LINKED ISSUE #91: Update PWA Manifest to indicate supported file types
Need to add something like this so that once the PWA is installed it will be offered as an option when selecting default file opener.
```json
"file_handlers": [
    {
      "action": "/",
      "accept": {
        "model/splat": [".splat"],
        "model/ply": [".ply"]
      }
    }
  ]
```

https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/How_to/Associate_files_with_your_PWA

willeastcott:
Now, when you install the SuperSplat PWA, PLY files will be associated with the app. Here is what that looks like on Windows:

pwaopen2

In the video about, I already have Paint3D associated with .PLY, so I use 'Open With' instead. I _could_ switch the SuperSplat PWA to always open .PLY instead, of course.

Fixes #91

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,9 +1,9 @@
 {
-    "name": "super-splat",
+    "name": "supersplat",
     "version": "0.16.0",
     "author": "PlayCanvas<support@playcanvas.com>",
-    "homepage": "https://playcanvas.com/super-splat",
-    "description": "All the splat things",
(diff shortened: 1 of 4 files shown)


slimbuck (approved):
<3

PR #105, opened 2024-06-06
Support loading multiple splats and use File System Api

slimbuck:
This PR:
* adds initial support for loading and editing multiple splats
* place file handling under file menu popup (see screenshot below)
* use File System Api for loading and saving files when available 
    * grandad Safari still uses the old file download path
* update to playcanvas [engine v1.71.5](https://github.com/playcanvas/engine/releases/tag/v1.71.5)
* bump package minor version

Screenshot 2024-06-06 at 15 30 54

### DIFF
--- a/src/controllers.ts
+++ b/src/controllers.ts
@@ -66,8 +66,9 @@ class MouseController {
             return vec2.sub2(a, b).length();
         };
 
-        let closestP = null;
         let closestD = 0;
+        let closestP = new Vec3();
+        let closestSplat = null;
 
         for (let i = 0; i < splats.length; ++i) {
             const splat = splats[i] as Splat;
@@ -90,19 +91,18 @@ class MouseController {
                 // find intersection
                 if (plane.intersectsRay(ray, vec)) {
                     const distance = dist(vec, cameraPos);
-                    if (!closestP) {
-                        closestP = vec.clone();
+                    if (!closestSplat || distance < closestD) {
                         closestD = distance;
-                    } else if (distance < closestD) {
                         closestP.copy(vec);
-                        closestD = distance;
+                        closestSplat = splat;
                     }
                 }
             }
         }
 
-        if (closestP) {
+        if (closestSplat) {
             this.camera.setFocalPoint(closestP);
+            scene.events.fire('selection', closestSplat);
         }
     };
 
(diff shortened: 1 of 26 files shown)


src/controllers.ts
```diff
@@ -24,6 +26,7 @@ import {
 const plane = new Plane();
 const ray = new Ray();
 const vec = new Vec3();
+const vec2 = new Vec3();
```
willeastcott, line 29: Gotta say, this line confused me to begin with...

slimbuck: I had actually thought about making them vecA and vecB because I thought it might look confusing, but then decided I was too lazy :)

slimbuck: I renamed it vecb

src/style.scss
```diff
@@ -69,10 +69,41 @@ body {
     #keyboard-panel & { content: '\E136'};
 }
 
-#file-selector-container {
-    flex-shrink: 0;
-    flex-grow: 0;
-    margin: 5px;
+#scene-panel-splat-list-container {
+    // padding: 10px;
```
willeastcott, line 73: Delete?

willeastcott (approved):
A superb enhancement! 👏

src/style.scss
slimbuck: done

CiberNin:
IMO there should be a little bit of vertical space between the file icon and the individual tool icons. Possibly even a slightly contrasting horizontal line.  Just as a little extra hint to users because file handling functionality is not particularly related to individual editor tools. Prep for potential future where you want to pack several groups of buttons in the toolbar.

slimbuck:
Yep, I agree @CiberNin.

I actually tried quickly adding some space between the file icon and tool icons, but it looked odd so I left it for now.

The whole UI needs an overall imo, which we'll hopefully get to soon.

slimbuck:
Although - if you have the time, feel free to tweak the CSS and submit a PR! Perhaps a little darkened divider and consistent spacing will work great.

attilaolah:
This works very well! What would also be really cool is to have a more precise way of orienting multiple splats, e.g. by selecting a few points that appear in both scenes.

slimbuck:
Hey @attilaolah , if you have the time please create an issue with this request (including details on how you think this should work). You never know, we may have spare cycles at point to implement it. :)

Issue #118, opened 2024-06-28
bug： Cropping without depth testing


Bin-ze:
When cropping the scene, not only the gs points in front of the camera are cropped, but also the gs points behind the camera. This is unreasonable, so I think it is necessary to add depth judgment and crop the point cloud in front of the camera.

slimbuck:
Hi @Bin-ze ,

Oh my gosh - this is a bug!

Will fix ASAP, thanks for reporting.

Bin-ze:
@slimbuck  I want to know when the web version will be updated？ 

https://playcanvas.com/supersplat/editor : There is still a problem with the Gaussian after cropping the camera

willeastcott:
@Bin-ze SuperSplat should be updated early next week.

PR #119, opened 2024-06-28
Rect select fix

slimbuck:
Fixes: #118

This PR:
* fixes a bug where points behind the camera were being selected in the rect tool in centers mode
* converts rect, brush and point tool to use pointer events
* capture pointer during drag so button release is correctly handled
* update to latest [Engine v1.72.0](https://github.com/playcanvas/engine/releases/tag/v1.72.0)
* bump package version

### DIFF
--- a/src/tools/picker-selection.ts
+++ b/src/tools/picker-selection.ts
@@ -9,7 +9,7 @@ class PickerSelection {
         this.root.id = 'select-root';
 
         this.root.addEventListener('pointerdown', (e) => {
-            if ((e instanceof MouseEvent) ? e.button === 0 : e.isPrimary) {
+            if (e.pointerType === 'mouse' ? e.button === 0 : e.isPrimary) {
                 e.preventDefault();
                 e.stopPropagation();
 
--- a/src/tools/picker-selection.ts
+++ b/src/tools/picker-selection.ts
@@ -8,24 +8,24 @@ class PickerSelection {
         this.root = document.createElement('div');
         this.root.id = 'select-root';
 
-        this.root.onmousedown = (e: MouseEvent) => {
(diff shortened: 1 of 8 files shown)


src/tools/rect-selection.ts
```diff
@@ -24,66 +23,72 @@ class RectSelection {
...
+        root.addEventListener('contextmenu', (e) => {
             e.preventDefault();
-        };
+        });
 
-        root.onmousedown = (e) => {
-            e.preventDefault();
-            e.stopPropagation();
+        root.addEventListener('pointerdown', (e) => {
+            if (dragId === undefined && ((e instanceof MouseEvent) ? e.button === 0 : e.isPrimary)) {
```
willeastcott, line 45: This should always be a `PoinerEvent`. Shouldn't you do this instead:
```
e.pointerType === 'mouse'
```

src/tools/picker-selection.ts
```diff
@@ -8,24 +8,24 @@ class PickerSelection {
         this.root = document.createElement('div');
         this.root.id = 'select-root';
 
-        this.root.onmousedown = (e: MouseEvent) => {
-            e.preventDefault();
-            e.stopPropagation();
+        this.root.addEventListener('pointerdown', (e) => {
+            if ((e instanceof MouseEvent) ? e.button === 0 : e.isPrimary) {
```
willeastcott, line 12: Again, maybe use `e.pointerType`?

willeastcott (approved):

src/tools/rect-selection.ts
slimbuck: yep that's simpler, thanks

PR #126, opened 2024-07-10
Add option to hide splats and remove them

slimbuck:
This PR:
- adds a visible toggle and remove button to the scene splat list
- only save visible splats

### DIFF
--- a/src/splat.ts
+++ b/src/splat.ts
@@ -139,6 +139,7 @@ class Splat extends Element {
     worldBoundStorage: BoundingBox;
     localBoundDirty = true;
     worldBoundDirty = true;
+    visible_ = true;
 
     constructor(asset: Asset) {
         super(ElementType.splat);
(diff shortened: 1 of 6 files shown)


mvaligursky (approved):

src/splat.ts
```diff
@@ -139,6 +139,7 @@ class Splat extends Element {
     worldBoundStorage: BoundingBox;
     localBoundDirty = true;
     worldBoundDirty = true;
+    visible_ = true;
```
willeastcott, line 142: What's with the trailing underscore? Don't we normally prepend it for private class fields?

willeastcott (approved):
Nice!

src/splat.ts
slimbuck: yes I guess. will fix in a followup PR. thanks!

PR #127, opened 2024-07-10
Add data panel

slimbuck:
Screenshot 2024-07-10 at 13 51 06

This PR:
* adds a new collapsable data panel below the 3d view
    * the view displays histogram for currently selected splat data
    * user can choose to plot one of the following properties:
        * `X`,  `Y`,  `Z`,  `Volume`,  `Surface Area`,  `Scale X`,  `Scale Y`,  `Scale Z`,  `Red`,  `Green`,  `Blue`,  `Opacity`,  `Hue`,  `Saturation`,  `Value`
    * click and drag on histogram to modify selection (using shift to add and ctrl to remove from selection)
    * hover mouse over histogram to see bucket values and counts.
* removed 'Select by opacity' and 'Select by Size' from control panel since histogram select is much more powerful
* bumped package version


src/ui/data-panel.ts
```diff
@@ -0,0 +1,372 @@
...
+    container.append(value);
+
+    parent.append(container);
+
+    return value;
+};
+
+class DataPanel extends Panel {
+    constructor(events: Events, args = { }) {
+        args = Object.assign(args, {
```
willeastcott, line 79: Modern JS would be:

args = { ...args,

src/ui/shortcuts-popup.ts
```diff
@@ -20,11 +20,12 @@ const shortcutList = [
     { header: 'SHOW' },
     { key: 'H', action: 'Hide Selected Splats' },
     { key: 'U', action: 'Unhide All Splats' },
+    { key: 'D', action: 'Toggle Data Panel' },
```
willeastcott, line 23: If you _do_ rename the panel, this would need updating too.

willeastcott (approved):
Can't wait to see what people think of this update!

src/ui/data-panel.ts
slimbuck: I've updated them all to use the _modern way_ ™

src/ui/data-panel.ts
```diff
@@ -0,0 +1,372 @@
...
+            headerText: 'Data',
```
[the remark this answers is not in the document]
slimbuck: Changed to DATA, thanks

PR #134, opened 2024-07-17
New grid, bounding boxes, perf improvements and more

slimbuck:
This PR:
* adds a bounding box indicator to the currently selected splat
* replaces the existing grid with an infinity-style version
* updates to the latest [engine v1.73](https://github.com/playcanvas/engine/releases/tag/v1.73.0)
* improves editor performance on large scenes when many splats have been deleted by removing the deleted splats from sorting and rendering
* finally addresses the issue where keyboard shortcuts sometimes don't work correctly
* bump package version

Screenshot 2024-07-17 at 15 21 11

### DIFF
--- a/src/scene.ts
+++ b/src/scene.ts
@@ -17,7 +17,8 @@ import { Model } from './model';
 import { Splat } from './splat';
 import { Camera } from './camera';
 import { CustomShadow as Shadow } from './custom-shadow';
-import { Grid } from './grid';
+// import { Grid } from './grid';
+import { InfiniteGrid as Grid } from './infinite-grid';
 
 class Scene {
     events: Events;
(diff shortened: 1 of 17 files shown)


src/scene.ts
```diff
@@ -17,7 +17,8 @@ import { Model } from './model';
 import { Splat } from './splat';
 import { Camera } from './camera';
 import { CustomShadow as Shadow } from './custom-shadow';
-import { Grid } from './grid';
+// import { Grid } from './grid';
```
willeastcott, line 20: Delete?

willeastcott: And can `grid.ts` be deleted too?

willeastcott (approved):
Really great PR - love it!

src/scene.ts
slimbuck: We can delete these. I kept them incase users would like the option of old grid or incase we need to revert this one. I'll remove the code once we're happy to keep the new grid.

Issue #70, opened 2024-02-15
Position Rotation Scale - missing SCENE tab


arlo-ml:
hi,
is there a reason why with the new version we can't rotate the file inserting the values in the SCENE tab?
sometimes you just need to rotate 90 degrees, and you can't have a good control using transform gizmos 

![supersplat](https://github.com/playcanvas/super-splat/assets/63562596/1eee99a4-ce4a-4fb0-9e59-f005dac362a1)

showpixelVR:
> hi, is there a reason why with the new version we can't rotate the file inserting the values in the SCENE tab? sometimes you just need to rotate 90 degrees, and you can't have a good control using transform gizmos
> 
[...]

Issue #110, opened 2024-06-12
[Feature Request] Precise rotation


attilaolah:
I happen to have many PLY files (splats) that are rotated e.g. 90° or 180° around a single axis. I didn't find a way to rotate the entire scene by a precise amount.

slimbuck:
Fixed in #143

PR #143, opened 2024-07-23
Transform panel

slimbuck:
Fixes: #70, #110, 

This PR:
* adds back the transform panel for editing splat position/rotation/scale
* select splat when toggling visibility
* fix small camera fly bug
* bump package version

Screenshot 2024-07-23 at 15 47 59

### DIFF
--- a/src/ui/splat-list.ts
+++ b/src/ui/splat-list.ts
@@ -1,0 +1,129 @@
+import { Container, Label, Element as PcuiElement } from 'pcui';
+
+class SplatItem extends Container {
+    getSelected: () => boolean;
+    setSelected: (value: boolean) => void;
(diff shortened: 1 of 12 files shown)


src/ui/splat-list.ts
```diff
@@ -0,0 +1,129 @@
+import { Container, Label, Element as PcuiElement } from 'pcui';
```
willeastcott, line 1: Why is `Element as PcuiElement` even necessary?

willeastcott (approved):

src/ui/splat-list.ts
slimbuck: It was necessary when SplatList was defined in control-panel.ts, no more.

Updated - thanks!

PR #194, opened 2024-09-25
Add selection updates to undo history

slimbuck:
This PR:
* places selection edits in the undo history
* pivot placement is also added
* render debug overlay from splat data directly (instead of building and updating separate copy of data)

### DIFF
--- a/src/tools/transform-tool.ts
+++ b/src/tools/transform-tool.ts
@@ -1,9 +1,8 @@
-import { TransformGizmo, Vec3 } from 'playcanvas';
+import { Entity, GraphicsDevice, TransformGizmo, Vec3 } from 'playcanvas';
 import { Scene } from '../scene';
 import { Splat } from '../splat';
 import { Events } from '../events';
-import { EditHistory } from '../edit-history';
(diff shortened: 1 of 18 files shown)


src/tools/transform-tool.ts
```diff
@@ -13,173 +12,145 @@ const patchGizmoMaterials = (gizmo: TransformGizmo) => {
...
-        this.splat = splat;
-        this.oldPivot = oldPivot;
-        this.newPivot = newPivot;
-    }
-
-    do() {
-        this.splat.setPivot(this.newPivot);
+// set the gizmo size to remain a constant size in screen space.
+// called in response to changes in canvas size
+const updateGizmoSize = (gizmo: TransformGizmo, device: GraphicsDevice) => {
```
willeastcott, line 17: Is this function still needed?

willeastcott (approved):
Nicely done!

src/tools/transform-tool.ts
slimbuck: on engine v1, yes

PR #206, opened 2024-10-04
german-l10n

simonbethke:
I just did a translation of all the strings to german language. During testing I found that one header wasn't localized yet so I did that one too in german and english.


willeastcott (approved):
Wunderbar! Thanks so much!

PR #207, opened 2024-10-06
Compress-selection

simonbethke:
This is an improvement to shrink the memory footprint of the undo-history.
What this basically does is, that a indices are combined as ranges if they cover more than one subsequent. So this:
1, 2, 3, 4, 5, 6, 10 
becomes:
1-6, 10

When selecting only few splats, this compression shows results in a typical range of 50 - 100% of the uncompressed size.

When selecting many splats like a tenth of the scene, it becomes more efficient and the typical compressed size is 2-20%.

When selecting almost all splats, the compressed size is less than 1%.

### DIFF
--- a/src/edit-ops.ts
+++ b/src/edit-ops.ts
@@ -23,24 +23,29 @@ const buildIndex = (total: number, pred: (i: number) => boolean) => {
     let rangeStart = -1;
 
     let saved = 0;
+    let predi = false, last = false;
 
     for (let i = 0; i < total; i++) {
-        if (pred(i)){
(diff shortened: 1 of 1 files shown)


simonbethke:
@slimbuck is this something you have already solved in a different way or is there something fundamental which you don't like on this implementation, but I should give it another shot from a different starting point?
If you don't want this for any reason, please just reject it. Else, what can I do to get this finished?

slimbuck:
This idea was eventually rolled into #803.

Closing, but with a little sadness to see it go :D

PR #211, opened 2024-10-11
tag exported ply files

simonbethke:
Similar to Nerfstudio, where I requested https://github.com/nerfstudio-project/nerfstudio/pull/3472 to write the term "nerfstudio" into the header, also super-splat should always do that.


src/splat-convert.ts
```diff
@@ -71,6 +71,7 @@ const convertPly = (convertData: ConvertEntry[]) => {
     const headerText = [
         `ply`,
         `format binary_little_endian 1.0`,
+        `comment generated by super-splat`,
```
willeastcott, line 74:
```suggestion
        `comment generated by supersplat`,
```

willeastcott:
Add SuperSplat version number?

simonbethke:
Yes, I was only choosing this string as it is used at another place this way. I'll clean it up

src/splat-convert.ts
simonbethke: I like the word to be written "SuperSplat" as it is in the titel of my browser-window. I took the hyphen-version only, because it exsted already for compressed ply

slimbuck (approved):
<3<3

src/splat-convert.ts
```diff
@@ -7,12 +7,15 @@ import {
...
 import { State } from './splat-state';
 import { SHRotation } from './sh-utils';
+import { version } from '../package.json';
 
 interface ConvertEntry {
     splatData: GSplatData;
     modelMat: Mat4;
 }
 
+const generatedByString =  `generated by SuperSplat ${version}`;
```
willeastcott, line 17:
```suggestion
const generatedByString =  `Generated by SuperSplat ${version}`;
```

willeastcott (approved):

PR #213, opened 2024-10-12
Tag-exported-ply-files

simonbethke:
Here is an addition:

Examples:

comment Generated by Nerfstudio 1.0
comment Vertical Axis: z

comment Generated by SuperSplat 1.0
comment Vertical Axis: y


simonbethke:
This is basically just the first step that is required, to rectify issues like wrong up-axis during loading of files. 
I filed basically the same PR for Nerfstudio and this would be the best answer for any issue anybody opens regarding "My file is opening with the wrong orientation, can you please fix it?"
And your answer should be: Contact the developer of your ply-producer and ask him to add the correct "comment Vertical Axis: "

slimbuck:
I'm not sure how useful this is going to be without adding support for all the possible Y-axis specifiers at load time. Won't the "Generated by" comments be enough for now?

slimbuck:
Unfortuntely the engine's resource loader doesn't expose comments otherwise you could detect this quite easily :(

slimbuck:
We rolled out custom header entries before, but unfortunately it broke some user's workflow with external tools that don't support any variation on the the PLY header.

I think the best solution is to add optional custom header bits for users that want them.

For that though we'll need a save and load dialog. So I'm closing this for now till we're ready to add that.

Issue #212, opened 2024-10-11
Value field manipulation with mouse doesn't work


simonbethke:
There is this kind of number-input that shows a left-right arrow when hovering to indicate that you can decrease the value by dragging it left and increase it by dragging it right. 
However, when modifying the position of a splat, you can drag left and it decreases, now drag right and it continues to decrease.
When creating a sphere for sphere-selection and you drag the radius field, the sphere size gets increased in both directions.

slimbuck:
@simonbethke could you please retest this? I believe this happened with me once, but haven't been able reproduce since. (I believe this was fixed with a recent gizmo update).

slimbuck:
Closing, but please reopen if you do find this still happens.

PR #217, opened 2024-10-16
Transform individual splats

slimbuck:
This PR add the ability to move/rotate/scale individual splats using transform gizmo and transform panel.

https://github.com/user-attachments/assets/716be0ff-7cf6-4dbe-86d5-897034fbd84b

The code changes can be summarised as follows:
- the transform panel and gizmos now interact with a `Pivot` object
- two new `TransformHandler` classes respond to pivot updates by applying the changes to the scene and updating the undo/redo history
- in order to transform large scenes in realtime we introduce a `Transform Palette`
    - the palette stores 3x4 transform matrices in an RGBA32F texture
    - each splat has a uint32 transform palette index assigned. 0 means no transform.
    - the transform palette is managed and updated during edits
- since splat positions are dynamic, some JS (cpu) processing has to be moved to GLSL (gpu), including:
    - intersection testing for mask/rect/sphere selections
    - bounding box calculations
    - splat centers updates (used for render sorting of splats)
    - these are all implemented in `data-processor.ts`

## Issues/Concerns
- this change results in higher memory usage (primarily for transform palette indices per splat) and slightly slower rendering due to transform palette
- the transform palette supports 4096 matrices initially and just doubles in size every time more space is needed
- the SPLAT DATA panel doesn't reference the transformed splat positions. this will be fixed in a followup PR.
- #212 is still present, will be addressed in a followup PR

## Testing notes
Full regression testing is needed, specifically:
- [ ] PLY and compressed PLY loads correctly
- [ ] PLY, compressed PLY and .splat formats save correctly with arbitrary edits
- [ ] gizmos work correctly under position/rotation/scale
- [ ] the transform panel works correctly under position/rotation/scale changes
- [ ] undo/redo history works correctly


src/camera.ts
```diff
@@ -310,10 +304,9 @@ class Camera extends Element {
         this.entity.camera.renderTarget = renderTarget;
         this.entity.camera.camera.horizontalFov = width > height;
 
-        // create pick mode render target
-        this.pickModeColorBuffer = createTexture(width, height, pixelFormat);
+        // create pick mode render target (reuse color buffer)
         this.pickModeRenderTarget = new RenderTarget({
-            colorBuffer: this.pickModeColorBuffer,
+            colorBuffer: colorBuffer,
```
willeastcott, line 309: You can shorthand this to just `colorBuffer`.

src/data-processor.ts
```diff
@@ -0,0 +1,399 @@
...
+        const { device } = this;
+        const { scope } = device;
+
+        const numSplats = splat.splatData.numSplats;
+        const transformA = splat.entity.gsplat.instance.splat.transformATexture;
+        const splatTransform = splat.transformTexture;
+        const transformPalette = splat.transformPalette.texture;
+
+        // update view projection matrix
+        const camera = splat.scene.camera.entity.camera.camera;
```
willeastcott, line 234: Do you need the final `.camera` here?

src/shaders/splat-shader.ts
```diff
@@ -0,0 +1,170 @@
...
+
+    vec4 v1v2 = calcV1V2(splat_cam.xyz, covA, covB, transpose(mat3(model_view)));
+
+    // get color
+    color = texelFetch(splatColor, splatUV, 0);
+
+    // calculate scale based on alpha
+    // float scale = min(1.0, sqrt(-log(1.0 / 255.0 / color.a)) / 2.0);
+
+    // v1v2 *= scale;
```
willeastcott, line 74: Any need to keep this?

simonbethke:
Saving is currently broken:
TypeError: Cannot read properties of null (reading '0')
    at uw.getMat (splat-convert.ts:74:40)
    at mw (splat-convert.ts:188:40)
    at file-handler.ts:333:28
    at file-handler.ts:330:22
    at async Sy.<anonymous> (file-handler.ts:225:13)

src/data-processor.ts
simonbethke: What is the semantic here?
splat -> the current
scene -> its scene
camera -> teh camera (so no more than one camera is supported in general? Orthographic projections could be helpful)
enity -> get the object properies of the camera (like transforms)
camera ?
camera ??

src/camera.ts
mvaligursky: typically I try to avoid it .. as then sometimes I rename the local variable and wonder why the code no longer works.

slimbuck: done, and elsewhere

src/data-processor.ts
slimbuck: > Do you need the final .camera here?

Nope, removed, thanks.

> What is the semantic here?

Yep, bit of a mouthful:
- splat - the SS splat instance
- scene - the SS scene instance
- camera - the SS camera
- entity - the playcanvas engine `Entity` instance containing transforms and scene/render hierarchy
- camera - the entity's camera component 
- camera (playcanvas engine'c camera instance, but component is enough so unneeded, removed)

src/shaders/splat-shader.ts
slimbuck: Yes, I might reintroduce this optimisation when I have a chance.

slimbuck:
> Saving is currently broken:

Thanks for this @simonbethke, fixed.

I hadn't tested _**not changing anything**_ then trying to save yet :)

simonbethke:
The "Reset Splat" function is not resetting the TransformsPalette

src/transform-palette.ts
```diff
@@ -0,0 +1,104 @@
...
+    free: (num?: number) => void;
+    texture: Texture;
+
+    constructor(device: GraphicsDevice, initialSize = 4096) {
+        let texture: Texture;
+        let data: Float32Array;
+
+        // reallocate the storage texture and copy over old data
+        const realloc = (width: number, height: number) => {
+            const newTexture = new Texture(device, {
```
mvaligursky, line 34: just a note that when you move this to the engine v2 you can use texture.resize()

mvaligursky (approved):

PR #219, opened 2024-10-17
Simple Color Adjustments

simonbethke:
This is a simple approach to adjust colors of splats.
image

The current UI is as simple as it gets and as long as I have no feedback for the way the adjustments are handled at the moment I would not start implementing the saving of that.

### DIFF
--- a/src/edit-ops.ts
+++ b/src/edit-ops.ts
@@ -192,6 +192,40 @@ class ResetOp extends StateOp {
     }
 }
 
+interface EntityColorAdjustment {
+    brightness: number
+    temperature: number
+    tint: number
+}
+
+class EntityColorAdjustmentOp {
+    name = 'entityColorAdjustment';
+
+    splat: Splat;
+    oldAdj: EntityColorAdjustment;
+    newAdj: EntityColorAdjustment;
+
+    constructor(options: { splat: Splat, oldAdj: EntityColorAdjustment, newAdj: EntityColorAdjustment }) {
+        this.splat = options.splat;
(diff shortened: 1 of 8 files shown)


mvaligursky:
Typically, a temperature is a setting useful for photography. For postprocessing is often more useful to just hava tint color, giving users more options. Something to consider.

simonbethke:
I totally agree, that temperature is useful for photogaphy. Thats why I integrated it here. :)

mvaligursky:
It's too limited option, typically used to color correct images. When people create splats, they have an option to color correct images ahead of time. Further color adjustments to splats are sometimes needed to fit the splat into their 3D (non photo based) environment, or apply artistic changes to it. This is not a photo color balancing any more.

simonbethke:
I am people and I implemented it as I really need it.

slimbuck:
This is awesome @simonbethke, thanks so much!

Could you make the 3 controls sliders instead of entry boxes?

I think the COLOR panel actually requires its own popup. This would be toggled by a new icon on the right toolbar (similar to VIEW OPTIONS panel).

I will actually be adding a new panel for CAMERA POSES next. I will do this in a way that should make adding the COLOR panel easy.

And I'm happy to start with the relatively simple controls, we can always extend them in future.

simonbethke:
Awesome! I like to add sliders, but I have to learn more about the availabile ui widgets

slimbuck:
This PR was awesome to get the ball rolling, thanks again. Closing in favour of #267

Issue #238, opened 2024-10-26
Is there a way we could render an image with a certain view?


Deaddawn:

slimbuck:
Hi @Deaddawn , 

Do you mean you want to download a PNG image of the current view or something else?

Downloading a png of the viewport would be easy to implement, though I guess we'd want to hide the grid/bounding box/selections etc.

Thanks

Deaddawn:
> Hi @Deaddawn ,
> 
> Do you mean you want to download a PNG image of the current view or something else?
> 
> Downloading a png of the viewport would be easy to implement, though I guess we'd want to hide the grid/bounding box/selections etc.
> 
> Thanks

Yes, would that be possible to tell us how to do this?

slimbuck:
I will add that today 👍🏻 (but without automatically hiding bits).

Deaddawn:
> I will add that today 👍🏻 (but without automatically hiding bits).

Thanks a lot! Does "hiding bits" means I can't hide grid or selections? If so, it will be a little inconvenience :joy:

slimbuck:
Yes you'll just have to hide the grid yourself before taking the screenshot for now, that's all.

Otherwise we must add options to the UI to 'include grid in screenshot' etc, which I'd rather not do right now.

slimbuck:
Btw the png will have transparent background, is that ok?

Deaddawn:
> Yes you'll just have to hide the grid yourself before taking the screenshot for now, that's all.
> 
> Otherwise we must add options to the UI to 'include grid in screenshot' etc, which I'd rather not do right now.

Understand, thanks.

Deaddawn:
> Btw the png will have transparent background, is that ok?

white bg would be great if possible 😂

Deaddawn:
God speed! Master Jedi

Deaddawn:
Sry to ask, I didn't find the screen shot button in the live version, should I run this locally?

slimbuck:
Yes code is merged, but is still to be published on playcanvas.com. That'll happen today or tomorrow most likely.

Use local version if you need it now.

Deaddawn:
> Yes code is merged, but is still to be published on playcanvas.com. That'll happen today or tomorrow most likely.
> 
> Use local version if you need it now.

I understand that background can not be adjusted in the current verson, would that be possible that you can tell me the code where i can just change the background into white. Thanks in advance

slimbuck:
We will add background color picker real soon.

[...]

PR #240, opened 2024-10-28
Add ability to save screenshot

slimbuck:
Fixes: #238

This PR:
- adds a menu option "Save Screenshot" which will download a PNG version of the viewport.
- use lodpng for compression since inbuilt browser compression can destroy images with alpha
- add automatic translations for the other languages:
    - french version is confirmed
    - if you have a chance, please check the German @simonbethke
    - anyone else for the rest :)

Since the backbuffer is generated with premultiplied alpha, we must apply a background color before compression. 

In future we should consider adding more screenshot options:
- save without background (i.e. semitrans) or specify a background color
- optionally remove the debug visuals (grid, overlay, bounding box) so user doesn't have to do this manually
- support video output of the currently loaded camera path


src/editor.ts
```diff
@@ -528,6 +528,60 @@ const registerEditorEvents = (events: Events, editHistory: EditHistory, scene: S
...
+        }
+
+        const arrayBuffer = await compressor.compress(
+            new Uint32Array(pixels.buffer),
+            texture.width,
+            texture.height
+        );
+
+        // construct filename
+        const filename = replaceExtension(selectedSplats()?.[0]?.filename ?? 'SuperSplat', '.png');
```
willeastcott, line 574: What will this generate if I have a PLY loaded from a URL?

willeastcott (approved):

src/editor.ts
slimbuck: Filename is just the file name excluding all the other bits. I've updated the code to do this properly, thanks!

Issue #215, opened 2024-10-15
Custom canvas background color


RobLinkA:
Some point clouds have low opacity, light colors or are close to the background, which often causes them to be ignored. I hope to add the background color customization of the entire canvas in the settings menu, so that it will be easier to clean up messy points.

slimbuck:
Just so you know, you can already set the background color using a URL parameter like so:

https://playcanvas.com/supersplat/editor?bgClr=0.1,0.2,0.3

PR #243, opened 2024-10-28
Add background color picker

slimbuck:
Fixes: #215

This PR:
- adds a color selector for the background to the VIEW OPTIONS panel

Screenshot 2024-10-28 at 15 28 34

Notes:
- the selected background color isn't currently saved. this should be done along with all the other settings
- the background color is used when taking screenshots
- the pcui color picker box is placed awkwardly and only hides itself when clicking on the parent panel

### DIFF
--- a/src/editor.ts
+++ b/src/editor.ts
@@ -548,7 +548,7 @@ const registerEditorEvents = (events: Events, editHistory: EditHistory, scene: S
             // construct the png compressor
             if (!compressor) {
                 compressor = new PngCompressor();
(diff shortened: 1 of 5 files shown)


willeastcott (approved):
Cool, but it's so weird how the picker is offset like that. Did this break it: https://github.com/playcanvas/pcui/pull/354

slimbuck:
Submitted https://github.com/playcanvas/pcui/pull/365 to fix the niggles

PR #246, opened 2024-10-29
Update to engine v2

slimbuck:
This PR updates SuperSplat to use PlayCanvas engine v2.2.0.

In addition, the following changes were made:
- move scss files into /scss folder and switch `@use` instead of `@import`
- split more shader code into /shaders
- update to eslint v9
- removed unused shadow class
- update to latest gizmo API 
- small fix to render target usage
- small fix to gizmo update logic


willeastcott (approved):

src/camera.ts
```diff
@@ -252,37 +251,51 @@ class Camera extends Element {
...
             const camera = entity.camera;
 
-            const inv = new Mat4().mul2(camera.projectionMatrix, camera.viewMatrix).invert();
-
             const set = (name: string, vec: Vec3) => {
                 device.scope.resolve(name).setValue([vec.x, vec.y, vec.z]);
             };
 
+            // get frustum corners in world space
+            const points = camera.camera.getFrustumCorners(-100);
```
willeastcott, line 276: Fine but this is not public API...

src/camera.ts
```diff
@@ -359,12 +371,15 @@ class Camera extends Element {
         this.entity.camera.camera.horizontalFov = width > height;
```
willeastcott, line 371: Can't this just be:

```
this.entity.camera.horizontalFov = width > height;
```

slimbuck: yep thanks!

src/camera.ts
```diff
@@ -391,7 +406,7 @@ class Camera extends Element {
         this.fitClippingPlanes(this.entity.getLocalPosition(), this.entity.forward);
 
         const { camera } = this.entity;
-        camera.orthoHeight = 0.5 * this.distanceTween.value.distance * this.sceneRadius / this.fovFactor * (camera.camera.horizontalFov ? this.scene.targetSize.height / this.scene.targetSize.width : 1);
+        camera.orthoHeight = this.distanceTween.value.distance * this.sceneRadius / this.fovFactor * (this.fov / 90) * (camera.camera.horizontalFov ? this.scene.targetSize.height / this.scene.targetSize.width : 1);
```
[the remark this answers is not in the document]
slimbuck: nope, thanks.

src/camera.ts
slimbuck: I'm ok with that :)

PR #258, opened 2024-11-04
Add color pickers

slimbuck:
This PR:
- adds three more color pickers for selected color, unselected color and frozen color.
- splat shader is updated to support the configurable colors and is also simplified in the process

Screenshot 2024-11-04 at 11 51 33


willeastcott (approved):
Cooooooool. 👌

Issue #76, opened 2024-03-23
Lasso Selector 


showpixelVR:
It would be great for more precise selection of points along straight surfaces, to be able to use a

- Lasso-Tool
- Polygonal Lasso Tool <-  more important

![image](https://github.com/playcanvas/super-splat/assets/78081450/1e50f613-ac04-4690-b9d9-65fd492d3cf9)

slimbuck:
Done in #269

PR #261, opened 2024-11-05
Polygon Selection

simonbethke:
I just added a little Polygon selection tool to the toolbar.

https://github.com/user-attachments/assets/4f6c39bd-540f-4467-826e-b4224f8eed60


simonbethke:
Important things to test for QA are:
* Close with double click and by clicking the start
* When finishing the selection by clicking the start again the dashed line gets brighter
* Change tool with unfinished polygon and get back to this tool (the line should start from the beginning)
* Changing the view is locked as long as I am editing a polygon but works with the tool selected as long as no poly is made
* Regression: Adding (+Ctrl) and Removing (+Shift) the selection from other selections should work too
* The tool can be accessed with shortcut key "O". (I wish I could have used "P", but thats already the word "Picker" that is used as synonym for "Rectangle".... I don't know why)

simonbethke:
Question:
* If when closing, the last point is not used as selection-point, should the dashed line also snap to that start point during edit?

slimbuck (approved):
Amazing. I just tested this and it works great! Thanks so much!

A few notes:
- The double-click to close is a great addition. Wondering how we can let users know about it.
- 4px to close felt to me a little small. I'd make it bigger... maybe 8px?
- I'd quite like backspace to delete the previously placed point (I can can do this later).
- The mask canvas element should be shared with brush tool (I will do this later).

Thanks again!

willeastcott (approved):
I think this wins 'PR of the week' competition. 🎉

willeastcott:
I note this partially addresses #76. That issue states that polygon selection is actually more important than lasso selection. But we should probably leave it open assuming we also intend to add a lasso select.

PR #268, opened 2024-11-06
Reorganization of the selection-classes

simonbethke:
I just re-added the simple lasso-selection. I will keep this as draft right now as I need to extract a base-class for different selection-types. For doing that I also have to make actual typescript classes from that without defining all methods as local variables in the closure.


slimbuck:
I'd suggest you submit the lasso tool implementation separately from an attempt to rearchitecting the code.

simonbethke:
After adding the lasso-tool, i started extracting a base-class for all selection-tools. The simple lasso-selection now is handled here:
https://github.com/playcanvas/supersplat/pull/269

slimbuck:
Not planning on this reorg.

PR #271, opened 2024-11-07
Allow to change the brush-size with the mouse-wheel

simonbethke:
I implemented a hidden but useful feature that allows to adjust the brush-size with the mouse-wheel by holding down the shift-key.


slimbuck:
Please could you merge main into this PR then see a few changes I prepared here? https://github.com/slimbuck/supersplat/tree/brush-wheel-fixes

simonbethke:
I don't know what exactly you mean by merging main into this PR. What I can see in that PR is, that you are accessing a "shiftKey" flag on a WheelEvent that doesn't have this flag. I had more or less your version of the code and it wasnt working, because of that.

simonbethke:
My suggestion would be, to track the keydown/keyup events in the controller.ts or at any other more global space and query that to know the modifyer-key state.

slimbuck:
I'm not sure why you wouldn't see shiftKey tbh https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/shiftKey. It's definitely there.

Pressing shift key switches the wheel axis from Y to X using an actual mouse wheel and switches the X/Y axes on a touchpad too. Perhaps you were testing with a touch pad and that's why it appeared to work for you?

If you test your PR with an actual mouse you should also see that deltaX is being sent in the wheel message, no longer deltaY. (On my mac pad I have to swipe left/right not up/down when shift is pressed.

> I don't know what exactly you mean by merging main into this PR.

The main branch now has commits that you've not merged into this of yours branch. I was suggesting you merge the latest commits from main branch into this one.

simonbethke:
Ok, I didn't know that the axis are flipped when holding shift.

slimbuck:
Also, in the updated PR, I use the existing larger/smaller logic which correctly tests radius min/max size (unlike your implementation).

simonbethke:
Thanks for the fixes. I merged main and your branch. Then I tested and found that for my normal mouse in chrome on a windows computer the delta-axis doesn't change when holding shift. So I calculated deltaX + deltaY and worked with that value. There used to be also a property called wheelDelta, but it seems as if it doesn't exsit anymore. Maybe they changed the API a bit and maybe since then, the axis aren't changed anymore by holding shift.

slimbuck (approved):
I like your idea of combing both X and Y.

It's still not working quite right, but I'll address the remaining issues after merge.

PR #279, opened 2024-11-08
Persisting Settings Values

simonbethke:
This is the most simplistic persistence I could tihnk of. It is just listening to interesting events to store their values in localStorage. When initializing it is checking if those values already exist and replays them.


slimbuck:
Closing in favour of #389.

PR #283, opened 2024-11-12
Typesafe Colmap and Nerfstudio camara-pose loading

simonbethke:
I am currently looking into pose-loading and want to continue by support loading nerfstudio-data poses. However, for that the first step is to know what kind of json I am looking at. 
So I improved the type-safetyness of loading colmap poses. Each pose now is casted to an interface that provides named access to other (currently unused properties). Before that, every entry is checked to be well-formed. Arrays with 2 or 4 elements when 3 were required are rejected. The entire pose then will not be loaded (but all others are). 
Counting the number of poses is done after filtering invalid ones (this was a tiny issue in the old code)

I expanded this to also loading nerfstudio data.
Here is an example of nerfstudio-data:
[transforms.json.zip](https://github.com/user-attachments/files/17723606/transforms.json.zip)
[bike2.ply.zip](https://github.com/user-attachments/files/17723607/bike2.ply.zip)

The data-format is explained here: https://docs.nerf.studio/quickstart/data_conventions.html

### DIFF
--- a/src/file-handler.ts
+++ b/src/file-handler.ts
@@ -113,7 +113,7 @@ const toColmapPose = (pose: any) => {
 };
 
 const toNerfstudioPose = (pose: any) => {
-    if(!('transform_matrix' in pose) || !isNestedNumberArray(pose.position, 4, 4))
+    if(!('transform_matrix' in pose) || !isNestedNumberArray(pose.transform_matrix, 4, 4))
         return null;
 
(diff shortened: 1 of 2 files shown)


simonbethke:
Here are a few input images, that should show if the poses have been imported correctly.

frame_00001
frame_00080
frame_00327

simonbethke:
I think this is not helpful at the moment.

PR #313, opened 2024-12-06
Add spherical harmonics to compressed.ply export

slimbuck:
This PR:
- adds support for exporting spherical harmonics to compressed.ply format
- data is simple quantising to 8-bits per coefficient
- PlayCanvas engine v1.76 and v2.3 support this format


willeastcott (approved):

slimbuck:
Yep sorry about that, busy working on the fix now!

Issue #329, opened 2024-12-11
[Feature request] Support saving splat colors in URL parameters


RobLinkA:
I noticed that background color can be set via URL parameter (e.g., `bgClr=0.0,0.0,0.0`). It would be very useful to have similar URL parameters for:
- Selected splat color
- Unselected splat color

This would allow users to:
1. Save their preferred color configurations in browser bookmarks
2. Share specific visualization configurations via URLs
3. Maintain consistent settings across sessions without manual reconfiguration

Is this feature already supported? If so, could you please document the parameter names? If not, would you consider adding this functionality?

Current known URL parameter:
- Background color: `bgClr=r,g,b`

Related use case: When working with multiple models, I often need specific color configurations for better visualization. Being able to save these settings in bookmarks would significantly improve workflow efficiency.

slimbuck:
Hi @RobLinkA ,

I can add this quite easily, but of course this solution is a band-aid.

We have plans afoot to implement a document format which will store all settings for a session, including these.

See #324 for details and feel free to comment there.

BTW any setting that appears in the [scene config object](https://github.com/playcanvas/supersplat/blob/main/src/scene-config.ts#L6) can be set via URL parameter. For example, to disable the debug overlay you use `&camera.debug=false`.

slimbuck:
Checking the current scene config object I see (quite embarrassingly) that there are some legacy settings we must remove! `hideLeftShoe`?!? :D

RobLinkA:
That's so nice! Thank you!

PR #334, opened 2024-12-12
Specify colors by url parameter

slimbuck:
Fixes: #329

This PR:
- adds ability to overwrite `selectedClr`, `unselectedClr` and `lockedClr` the way `bgClr` can
- colors are specified as an array, for example `&selectedClr=0.8,0,0,0.5`
- also deleted some legacy classes from config and code


willeastcott:
Camel-case in URL params bothers me for some reason. But you're presumably doing this because they match up with JS property names?

slimbuck:
Yep

willeastcott (approved):

willeastcott:
I mean, don't want to be annoying here, but I'd almost prefer a `Map` of nice URL params to internal property names.

slimbuck:
We could do, but:
- this feature shouldn't be needed in the near future when we introduce document format
- if we have a map we'd  have to document it somewhere. currently there is just one reference and everything shares it

willeastcott:
OK, I shall bow to your supreme authority on this occasion.

slimbuck:
Which format would you prefer for url params, kebab-case? I might be able to do an auto-case-conversion...

willeastcott:
Yeah, I ❤️ kebabs. 😋

slimbuck:
I've added support for all values to specify either case, `bg-clr` or `bgClr` 👍🏻

PR #350, opened 2024-12-17
Rings mode picker update

slimbuck:
When picking gaussians in rings mode, only the top layer of gaussians were ever considered for adding/removing/setting the selection.

This has meant it is not possible to select anything other than the top layer of gaussians. Also, it has not been possible to deselect gaussians that are embedded within the scene, since the top layer of gaussians block picking.

This PR updates the picking logic to be much more useful and behave more consistently:
- when adding to the selection (shift), only consider unselected gaussians
    - this means it's now possible to add gaussians to the current selection layer at a time
- when removing from the current selection (ctrl), only consider the currently selected gaussians
    - so unselecting "blocked" gaussians is now possible
- when setting the selection, do as before

For example, this was not possible before:

https://github.com/user-attachments/assets/91df1a59-921f-4233-bd81-46bfe7ef6900

### DIFF
--- a/src/camera.ts
+++ b/src/camera.ts
@@ -486,7 +486,7 @@ class Camera extends Element {
         for (let i = 0; i < splats.length; ++i) {
             const splat = splats[i] as Splat;
 
-            this.pickPrep(splat);
+            this.pickPrep(splat, 'set');
             const pickId = this.pick(sx, sy);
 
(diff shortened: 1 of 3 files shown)


willeastcott (approved):
Hugely worthwhile improvement! 🚀

PR #354, opened 2024-12-18
Render grid in ortho view

slimbuck:
Screenshot 2024-12-18 at 11 05 24


willeastcott (approved):
This is so useful!

simonbethke:
I have one more request: can the colored 0-lines be always at least one or even two pixel wide? Often I have my object far off the 0-point and when trying to move it there, I either see the object or the colored lines...

PR #361, opened 2024-12-30
Fix undo/redo keyboard shortcut

### LINKED ISSUE #232: Undo/Redo causes unwanted rotation
Steps to reproduce:

1) Open a scene
2) Rotated it by typing a number in the rotation
3) Make a rectangular selection
4) Delete it using the del-key
5) Undo that using CTRL-Z
6) Press CTRL-Z again (no effect)
7) Now click the Undo-Button (the selection will rotate shortly and then rotated back)
8) Click the Redo-Button (you have a rotated selection)

slimbuck:
Fixes #232

This PR adds an option to register a keyboard event for capture phase (instead of the default bubble phase). Capture phase handlers are fired first, before any child listeners, allowing global overrides of events.

This is useful for undo/redo which is also handled by entry boxes and results in confusing undo/redo behaviour depending on the currently focused dom element.

This PR also stops propagation of the recognised up and down events and also prevents the default event firing. This is required for capture phase handler, but is probably also desirable for bubble events. This change needs testing.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -83,8 +83,8 @@ const initShortcuts = (events: Events) => {
     shortcuts.register(['U', 'u'], { event: 'select.unhide' });
     shortcuts.register(['['], { event: 'tool.brushSelection.smaller' });
     shortcuts.register([']'], { event: 'tool.brushSelection.bigger' });
-    shortcuts.register(['Z', 'z'], { event: 'edit.undo', ctrl: true });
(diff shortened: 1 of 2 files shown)


willeastcott (approved):

simonbethke:
Has this already been released? I ran into random UNDO-issues yesterday.

willeastcott:
@simonbethke No, I believe we are aiming to deploy the fix either today or tomorrow.

PR #364, opened 2025-01-03
Animation preview

slimbuck:
Add the ability to view a set of PLY files as an animation.

Screenshot 2025-01-03 at 13 52 25

In this PR:
- animation mode is enabled if user loads a set of files that are detected to form an animation sequence
    - user can either drag in multiple files or a folder containing multiple files
- upon detection the timeline panel is shown
- then the user can:
    - scrub the timeline
    - play and stop the animation
    - advance one frame forward and backward
    - set playback speed

This is very much work in progress and meant only for viewing animation sets. Don't attempt to load animations and non-animations and edit a scene at the same time.

Current issues:
- Each frame of the PLY animation is loaded as usual. This means you can't do anything about badly oriented scenes.

### DIFF
--- a/src/ui/scrubber.ts
+++ b/src/ui/scrubber.ts
@@ -1,0 +1,130 @@
+import { Container, Label, SelectInput, SliderInput } from 'pcui';
+import { Events } from '../events';
+import { Tooltips } from './tooltips';
+
+class Scrubber extends Container {
+    constructor(events: Events, tooltips: Tooltips, args = {}) {
+        args = {
+            ...args,
+            id: 'scrubber',
+            hidden: true
+        };
+
+        super(args);
+
+        const controlsContainer = new Container({
+            id: 'controls-container'
+        });
+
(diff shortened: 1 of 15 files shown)


src/ui/scrubber.ts
```diff
@@ -0,0 +1,131 @@
+import { Container, Label, SelectInput, SliderInput } from 'pcui';
+
+import { Events } from '../events';
+import { Tooltips } from './tooltips';
+
+class Scrubber extends Container {
```
willeastcott, line 6: I'd probably called this `Timeline` but hey.

willeastcott (approved):
So coooooool!

willeastcott:
Ideally, please update the Wiki with a section on animations.

src/ui/scrubber.ts
slimbuck: I was being a little lazy, but I've renamed thanks!

PR #373, opened 2025-01-09
Export updates

slimbuck:
This PR updates the export pipeline, specifically:
- update export settings structure layout and move logic to dialog handler
- unify the gaussian filter to support filtering of small and completely-transparent gaussians at export
- remove legacy material handling
- move export viewer settings to json instead of patching html
- update localisations
- add initial publish UI


src/templates/viewer-html-template.ts
```diff
@@ -238,7 +234,15 @@ const template = /* html */ `
...
                 }
 
-                entity.script.create(FrameScene);
+                const appElement = await document.querySelector('pc-app').ready();
+                const cameraElement = await document.querySelector('pc-entity[name="camera"]').ready();
+
+                const app = await appElement.app;
+                const camera = cameraElement.entity;
+
+                camera.camera.clearColor = new Color(viewerSettings.background.color);
```
willeastcott, line 243: You should be able to do just:

```
cameraElement.clearColor =
cameraElement.fov =
```

src/ui/localization.ts
```diff
@@ -25,7 +25,7 @@ const localizeInit = () => {
                     'scene.export': 'Exportieren',
                     'scene.export.compressed-ply': 'Komprimiertes PLY',
                     'scene.export.splat': 'Splat',
-                    'scene.export.viewer': 'Viewer App',
+                    'scene.export.viewer': 'Viewer App...',
```
willeastcott, line 28: What's with the `...`?

willeastcott (approved):

src/ui/localization.ts
slimbuck: shows a dialog will appear upon selection.

src/templates/viewer-html-template.ts
slimbuck: the next line sets .script on the same object, which isn't supported by camera web element.

Issue #358, opened 2024-12-21
[BUG] HTML export does not correctly keep viewport


pablovela5620:
Here's a video showing what I mean.

https://github.com/user-attachments/assets/7ff3f483-5848-4eac-85eb-07de43c8c8fa

Here is the original splat + HTML viewer files

https://huggingface.co/datasets/pablovela5620/supersplat-regression/blob/main/splat-viewer.html

slimbuck:
I've noticed that the browser sometimes caches old versions of the html file. Can you please try clearing the browser cache and then loading the html again?

pablovela5620:
I cleared browser data using 
```
Click the three vertical dots (⋮) in the top-right corner.
Navigate to More Tools > Clear browsing data.
In the Time range dropdown, select All time.
Check Cached images and files.
Click Clear data.
```
and I'm still getting the same results

pablovela5620:
I've tried both on my linux machine, and on my MacBook. Exporting on either `current viewport` or `1st camera` results in the broken view above

slimbuck:
Oh damn I know what the issue is. (Sorry just got a chance to look).

We set the camera's front and back clipping plane to cover the whole scene's extents and in this case it means the front clipping plane is too large.

You can edit the .html file and set line 203 and 204 something like:
```
this.entity.camera.farClip = 100;
this.entity.camera.nearClip = 0.01;
```
Sorry about this, it's actually come up once before and I just forgot to submit a fix.

pablovela5620:
Looks like that fixed it! Ill leave this issue for now until there's a more permanent fix

PR #376, opened 2025-01-10
Viewer html fix

slimbuck:
Fixes: #358

Instead of basing camera clipping plane on scene size we set static near and far planes.

### DIFF
--- a/src/templates/viewer-html-template.ts
+++ b/src/templates/viewer-html-template.ts
@@ -196,8 +196,8 @@ const template = /* html */ `
 
                         // configure camera
                         this.entity.camera.horizontalFov = true;
-                        this.entity.camera.farClip = bbox.halfExtents.length() * 20;
-                        this.entity.camera.nearClip = this.entity.camera.farClip * 0.001;
+                        this.entity.camera.nearClip = 0.01;
+                        this.entity.camera.farClip = 1000;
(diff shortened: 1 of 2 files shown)


mvaligursky (approved):
what if the splat is larger?

slimbuck:
Gaussians aren't clipped to near and far planes (see [shader](https://github.com/playcanvas/engine/blob/main/src/scene/shader-lib/chunks/gsplat/vert/gsplatCenter.js#L19)), so the se planes affect:
- camera culling of splats and meshes
- combining splats with non-splat meshes

I think the fixed planes will cover almost all cases.

willeastcott (approved):

src/templates/viewer-html-template.ts
```diff
@@ -196,8 +196,8 @@ const template = /* html */ `
 
                         // configure camera
                         this.entity.camera.horizontalFov = true;
-                        this.entity.camera.farClip = bbox.halfExtents.length() * 20;
-                        this.entity.camera.nearClip = this.entity.camera.farClip * 0.001;
+                        this.entity.camera.nearClip = 0.01;
+                        this.entity.camera.farClip = 1000;
```
willeastcott, line 200: Is it worth just setting these on the `pc-camera` then?

slimbuck: yes!

willeastcott (approved):

PR #379, opened 2025-01-15
Double click updates

slimbuck:
This PR:
- fixes a bug where gizmo is placed when double clicking, even when no gizmo was selected
- adds a 3d cursor label to the viewport which shows the location of the double click
    - click the text to copy it to clipboard

## Notes
- it would be awesome to extend this functionality so SuperSplat supports arbitrary markers with data attached

Screenshot 2025-01-15 at 12 03 05


willeastcott (approved):
Cool. Although maybe the double-click coordinate text element should be hidden once the user has clicked somewhere some time after the double-click.

PR #380, opened 2025-01-15
HTML Viewer update

slimbuck:
This PR:
- splits out the viewer's css and script into separate files for the .zip export option
    - this layout should make it easier for users to maintain
    - we can more easily add new scripts for various functionality as we add it
- move all styling to css and dom creation to html

## Notes
- a followup PR will use this new structure to add new functionality, like double-click to focus and FPS camera.


willeastcott (approved):
Nice.

Issue #236, opened 2024-10-25
SuperSplat file format


simonbethke:
**User story:**
As user of SuperSplat I would like to open and save files that contain SuperSplat specific data.

**Details:**
It would be very helpful, if SuperSplat could come up with an own dataformat. This format could for example be a renamed zip-file without any compression (just as convenient container). As file-extension *.ssd would be funny "SuperSplat Data". The Zip-File would contain a main.json file and all splat-files that have been imported as ply files.
The main.json file should have these properties:
* one entry per splat-file with:
  * filename
  * transformation matrix (translation, rotation, scale)
  * visibility
  * color-edits
* editor-camera-setup (transformation + fov)

Issue #324, opened 2024-12-10
High level document RFC


slimbuck:
### Background

SuperSplat has grown to include many settings which are currently not stored between editing sessions. This introduces friction for users who edit scenes over multiple sittings.

Fixing this will require us to introduce a document storage format for saving the editor state between sessions.

This feature has already been requested by users, see #236.

The document will store the following types of data:
- The splat objects, including their state (deleted, selected, frozen), color tinting, transform
- The list of camera poses
- Viewport settings
- Export and import settings
- Other future settings

The document will primarily be stored and loaded from the user's local hard drive (and perhaps optionally browser local storage) much the way PLY files are currently handled.

Ideally the document would be stored in a user-accessible way:
- A PLY file for each splat object (as binary data)
- A single JSON blob for everything else (as text data)

### Alternatives

We have a few options with regards to how we would package such data.

## **1. Create a custom binary file format and store everything together**

This would likely be the least effort solution.

Cons: Users would have no access to the data outside of SuperSplat.

## 2. Store the document in separate files

This might be possible using the browser's [File System API](https://developer.mozilla.org/en-US/docs/Web/API/File_System_API).  User would grant SuperSplat access to a folder at load and save time.

Pros: files placed nicely on disk would allow very nice user access

Cons: File System API is not supported in Safari and even Chrome users are shown "scare screens" when requesting read/write access to folders. Also we will have no ability to associate a file type with SuperSplat PWA.

## 3. Zip files into one archive

This could result in useful user-side disk space saving. 

Cons: It might not be possible to stream data to a multi-file zip target in which case memory requirement at save time is doubled.

willeastcott:
Should there be global settings (persisting regardless of what scene is loaded) and project specific settings? Or should all settings be tied to a specific project?

What should the project file extension be? `.ssplat` maybe?

I would probably go with a ZIP format containing a .json and .ply and rename it to .ssplat. If a user really needed to access the contents, they could rename to .zip and extract. But generally, that should not be needed by a typical user.

simonbethke:
Maybe there could be the option to zip-files, but at the same time allow to load the files unzipped. 
If its a zip-file, its your 3rd option. However, in that 3rd option you woudl have a json-file or something that holds the semantic state. If that file was outside the zip, it could still be loaded either alongside with the respective ply-files or entirely standalone which would make the editor show a bounding-box instead of the splat file. So the user can load the splat-data afterwards (and also only the objects he needs)

[...]

PR #389, opened 2025-01-28
Native document format

slimbuck:
Fixes: #236
Related tickets: #324

This PR implements support for a native SuperSplat document.

The Scene menu has been updated to reflect document vs import/export:

Screenshot 2025-01-28 at 16 39 45

## Document contents
The document is a simple uncompressed zip archive containing the following files:
- document.json
- splat_0.ply
- splat_1.ply
- ...
- splat_n.ply

## document.json
Contains all editor state apart from the individual splats. This includes:
- version, number starting at 0 and increasing with schema changes
- view settings
- camera settings
- camera poses
- per-splat settings.

## splat_n.ply
Each PLY present in the document is similar to its exported counterpart with the following differences:
- color tinting is not applied, but rather color tint settings are stored in document.json
- the splat's world transform is not applied, but rather stored in document.json

## Example
Here is a sample document.

[example.super.zip](https://github.com/user-attachments/files/18573030/example.super.zip)

The file has been renamed to a .zip extension to allow attachment here.

## Notes
- we write the document using a custom zip writer implementation in order to stream the file out and lower memory requirements
- loading the document is done with JSZip, but this likely results in memory overhead and will need to be addressed in future
- even though SuperSplat UI currently supports a single set of camera poses, the document supports multiple (since we are likely to add support for multiple in the near future).
- it would be prudent to document the schema in document.json. This will be discussed and done in a followup PR.
- we currently output the document with .super extension because I like it, but this might change based on the success or not of bribery @willeastcott


willeastcott:
Can the menu be:

```
New
Open
---
Save
Save As
---
Import
Export
---
Save Screenshot
```

slimbuck:
> Any reason not to ZIP it? Otherwise, these things are gonna eat up quite a lot of disk space, right?

Yes. JSZip and the other libraries I looked at require a blob/array of all data before writing out the zip file. This would double the memory requirement during save. We're streaming the data out instead.

src/doc.ts
```diff
@@ -0,0 +1,300 @@
...
+import { serializePly } from './splat-serialize';
+import { localize } from './ui/localization';
+
+// ts compiler and vscode find this type, but eslint does not
+type FilePickerAcceptType = unknown;
+
+const SuperFileType: FilePickerAcceptType[] = [{
+    description: 'SuperSplat document',
+    accept: {
+        'application/octet-stream': ['.super']
```
willeastcott, line 15: image

slimbuck: Given there is already .ksplat and .splat, I think users will confuse .ssplat with a runtime format.

slimbuck: Actually I think that's the worst possible option.

willeastcott (approved):

src/doc.ts
slimbuck: Bribery didn't work, we're going with `.ssproj` extension.

willeastcott:
I would recommend renaming the 'Scene' menu to 'File'.

slimbuck:
> I would recommend renaming the 'Scene' menu to 'File'.

It's a big change, will do in a followup PR.

PR #390, opened 2025-01-28
Rename 'Scene' menu to 'File'

slimbuck:
Since file is more standard.


willeastcott (approved):
So satisfying! ❤️

willeastcott:
I've just realized....we should not forget to update the Wiki docs...

PR #394, opened 2025-01-30
Smooth camera animation

slimbuck:
This PR:
- adds cubic (hermite) spline interpolation for camera movement from poses

https://github.com/user-attachments/assets/8fa2c3a9-430a-49a4-8a97-29565d2dba9b

## Notes:
- with the new native project format (which stores camera poses) and camera interpolation, we can now:
    - offer the option of exporting animated movies
    - export camera spline to html viewer
    - improve the camera poses panel considerably
        - support multiple camera pose sets
        - transform pose sets with gizmos
        - support basic delete/insert/modify controls
        - display camera poses in the viewport
        - support selection of camera pose from 3d viewport
        - support for various camera input/export formats (especially nerfstudio format cc @simonbethke)
        - approximately another million things

### DIFF
--- a/src/anim/spline.ts
+++ b/src/anim/spline.ts
@@ -1,0 +1,108 @@
+class CubicSpline {
+    // control times
+    times: number[];
+
+    // control data: in-tangent, point, out-tangent
+    knots: number[];
+
+    // dimension of the knot points
+    dim: number;
+
+    constructor(times: number[], knots: number[]) {
+        this.times = times;
+        this.knots = knots;
+        this.dim = knots.length / times.length / 3;
+    }
+
+    evaluate(time: number, result: number[]) {
+        const { times } = this;
+        const last = times.length - 1;
+
+        if (time <= times[0]) {
+            this.getKnot(0, result);
+        } else if (time >= times[last]) {
+            this.getKnot(last, result);
+        } else {
+            let seg = 0;
+            while (time >= times[seg + 1]) {
+                seg++;
+            }
+            return this.evaluateSegment(seg, (time - times[seg]) / (times[seg + 1] - times[seg]), result);
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


src/anim/spline.ts
```diff
@@ -0,0 +1,108 @@
...
+                if (i === 0) {
+                    tangent = (points[idx + dim] - p) / (times[i + 1] - t);
+                } else if (i === times.length - 1) {
+                    tangent = (p - points[idx - dim]) / (t - times[i - 1]);
+                } else {
+                    // finite difference tangents
+                    tangent = 0.5 * ((points[idx + dim] - p) / (times[i + 1] - t) + (p - points[idx - dim]) / (t - times[i - 1]));
+
+                    // cardinal spline tangents
+                    // tangent = (points[idx + dim] - points[idx - dim]) / (times[i + 1] - times[i - 1]);
```
willeastcott, line 92: Delete?

willeastcott (approved):
This is great! I do wonder about whether cameras should be keyed on the timeline though. Undecided on that.

src/anim/spline.ts
slimbuck: Plan is to perhaps make this tangent calculation an option in future. So for now it's _documentation_.

slimbuck:
Ultimately a timeline would be the best. The camera pose panel builds the spline with consistent time steps, but the underlying spline implementation handles arbitrary times. So once we add timeline it should largely still _just work_.

willeastcott:
> Ultimately a timeline would be the best. The camera pose panel builds the spline with consistent time steps, but the underlying spline implementation handles arbitrary times. So once we add timeline it should largely still _just work_.

Then we can go with this but if there's time, consider squeezing in a _very_ basic upgrade to our 4DGS timeline control. Let's see how things go over the next couple of days.

PR #399, opened 2025-02-08
Viewer export of camera animation

slimbuck:
This PR makes it possible to export camera animations to the html viewer, specifically:
- update the viewer templates and use rollup-plugin-string to import them in the exporter code
- add support to viewer for showing a poster at load time
- update viewer styling
- update dependencies

Notes:

- a follow-up PR will add a timeline editor for more useful animation editing
- the publish and export dialogs share much logic, these should share some common base ideally
- the camera animation doesn't loop smoothly. this will be done at the same time as timeline


src/templates/spline.js
```diff
@@ -0,0 +1,108 @@
+class CubicSpline {
```
willeastcott, line 1: It does feel so strange to me that the PlayCanvas Engine itself doesn't provide comprehensive support for curves. Keep in the back of your mind a way for us to migrate this at some point to that repo, ideally.

src/templates/spline.js
```diff
@@ -0,0 +1,108 @@
...
+                m0 * (t * omt2) +
+                p1 * (t2 * (3 - twot)) +
+                m1 * (t2 * (t - 1));
+        }
+    }
+
+    // create cubic spline data from a set of control points to be interpolated
+    // times: time values for each control point
+    // points: control point values to be interpolated (n dimensional)
+    // tension: level of smoothness, 0 = smooth, 1 = linear interpolation
```
willeastcott, line 70: Any reason not to just use JSDoc? 😄

willeastcott (approved):
Very exciting - can't wait to play with this!!

src/templates/spline.js
slimbuck: Because nothing in the project is typedoc'd! Added the comments for when we do though!

PR #406, opened 2025-02-16
Export fix

slimbuck:
This PR updates the compressed export (and publish) to, by default:
- remove gaussians with alpha < 1 / 255 
- remove gaussians with NaNs or infinites in any of their properties

This is done because we are seeing scenes published which contain a large numbers of invalid gaussian data (in some cases nearly half the scene). These gaussians don't render, but take storage and download space and slow down rendering, significantly in some cases.


willeastcott (approved):

willeastcott:
Wouldn't it be better to report warnings/errors on load/import and then give option to fix?

slimbuck:
> Wouldn't it be better to report warnings/errors on load/import and then give option to fix?

Yes would be good to add this too. In some cases values like -Infinity might actually be a valid value, for example for scale. But also we support blind PLY data which user can attach and we pass through unchanged.

We will never want these values in a compressed.ply though, under no circumstance. So I think this step is still needed.

PR #410, opened 2025-02-19
[FIX] Limit accepted file types for Open and Import

willeastcott:
`File > Open`:

Before:

image

After:

image

`File > Import`:

Before:

image

After:

image


simonbethke:
I would even suggest to allow "opening" ply and splat.

slimbuck (approved):

PR #417, opened 2025-02-22
Do not reset camera clearColor in viewer AR mode

vincentwoo:
Currently the viewer does a strange thing in AR mode - it sets the camera clear color to fully transparent, which has the effect of showing you your physical environment in AR mode wherever there is transparency in your scene. This simply deletes this behavior. I don't have an intuition for why it exists previously but feel free to ignore this PR if it's essential for something. In testing it in my viewer app it seems to work fine (Pixel 7 Chrome AR mode).

### DIFF
--- a/src/templates/index.js
+++ b/src/templates/index.js
@@ -279,44 +279,6 @@ document.addEventListener('DOMContentLoaded', async () => {
         });
     }
 
-    // On entering/exiting AR, we need to set the camera clear color to transparent black
-    let cameraEntity, skyType = null;
-    const clearColor = new Color();
-
-    app.xr.on('start', () => {
-        if (app.xr.type === 'immersive-ar') {
-            cameraEntity = app.xr.camera;
-            clearColor.copy(cameraEntity.camera.clearColor);
-            cameraEntity.camera.clearColor = new Color(0, 0, 0, 0);
-
-            const sky = document.querySelector('pc-sky');
-            if (sky && sky.type !== 'none') {
-                skyType = sky.type;
-                sky.type = 'none';
-            }
-
-            app.autoRender = true;
-        }
-    });
-
-    app.xr.on('end', () => {
-        if (app.xr.type === 'immersive-ar') {
-            cameraEntity.camera.clearColor = clearColor;
-
-            const sky = document.querySelector('pc-sky');
-            if (sky) {
-                if (skyType) {
-                    sky.type = skyType;
-                    skyType = null;
-                } else {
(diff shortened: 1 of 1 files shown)


Maksims:
That is definitely intentional, and the whole point of AR mode is to see passthrough and to achieve that you need to set the camera's clear colour to transparent.
If you want a VR experience, then launch VR type session.

vincentwoo:
I see your point. It might be good to make this configurable as a viewer setting. The more background-encompassing a splat is, the more jarring it is to see your physical environment in the rendering. Another solution might be to optionally have a transparency falloff around the POI when in AR mode.

Maksims:
The most simple solution is to provide VR and AR buttons instead of a single XR one and let users choose which context they want.

vincentwoo:
both VR and AR buttons are present already, the problem is 99% of people have a phone and 1% of people have a VR headset. When you hit the VR button on a phone it defaults to a split screen stereoscopic, which isnt a very good first-run experience.

Maksims:
So we could hide VR button on the mobile phones, and only offer AR button where it is supported. And on headsets offer both options, and disable unsupported session types (e.g. on Desktop and AVP, AR is not available).

Maksims:
> On mobile we could continue to offer AR and VR modes to the user. Both modes would work like the current AR mode, but we would clear the background in the case of VR mode.
> 
> Would this limit google cardboard-type experience. Does anyone still use this? (Or is there a modern alternative?)
> 
> @Maksims does this make sense? Can we "detect" mobile devices easily?

On mobile, offering a VR mode probably does not make much sense. Cardboard is not a thing anymore, and Samsung Gear VR is not either.
But Quest - is technically a mobile device too, and it does support well both, VR and AR, and has reasons to have both modes.
Detecting specifically if page is viewed from the stereo device - unfortunately you can only do that using userAgent, and there is no JS reliable way to do that.

When rendering using AR but clearing the background - does achieve pretty much same as VR, but has some performance drawbacks as it has to use some device resources for MR related things.
There was a [recent PR](https://github.com/immersive-web/webxr/pull/1404) to WebXR specs to allow easy disabling of MR features when fully obscuring viewport, like in VR, but it is up to implementers to catch up. I think this would be the best use case here: if AR supported offer only it, and allow easy toggle to solid or passthrough background.

slimbuck:
Closing this as development is now at https://github.com/playcanvas/supersplat-viewer.

We have an issue https://github.com/playcanvas/supersplat-viewer/issues/8 where I have mentioned this PR for reference.

PR #463, opened 2025-04-10
Very simple implementation of a color-temperature control

simonbethke:


slimbuck:
Nice one @simonbethke. I actually love this and feel like users will benefit from direct temperature control.

I'm wondering about the range of the temperature slider. Usually these are marked with the corresponding kelvin. Can we do the same?

slimbuck:
An issue that bugs me all the color controls is that they are applied in gamma space.

We do this because the spherical harmonics are stored in gamma space too.

The tints should be applied in linear space, but I will have to investigate whether this is even possible.

simonbethke:
Okay, there are 2-3 arguements which made me use a unitless control over actual Kelvin:
- Kelvin calculation requires log function which might be calculation intensive
- Normally for Kelvin calculation you know the starting point (for example 6500k) I have never seen a relative kelvin adjustment like -1000k / +1000k. I can imagine, that you cannot have an absolute +1000k if you don't know if you do 6000+1000 or 7000+1000. I don't want to label something "Kelvin" if its just wrong
- In the end, the user will put the hands on the slider and look what happens, so colorspace and the actual calculation doesn't matter too much.

simonbethke:
Grok answers this:

Without knowing the starting Kelvin value of your "unknown source," you can’t precisely adjust it by an absolute amount like +1000K in RGB alone. The adjustment depends on:

The starting color temperature (e.g., 3000K + 1000K = 4000K looks different from 6000K + 1000K = 7000K).
The RGB color space (e.g., sRGB, Adobe RGB), as each has different gamut limits and white points.

simonbethke:
I don't know if you know that I do hate gimp. I love photoshop and gimp tries to do everything photoshop does and it is really good in doing everything worse :D

I really don't see any use for units or accurate adjustments. I did plenty of photo adjusting in lightroom, but I never thought "I need to reduce the temperature by 50k"

slimbuck:
Well the slider is adjusting _something_ and showing a value, so if possible it might as well make sense.

slimbuck:
...in which case current units are fine

src/ui/localization.ts
```diff
@@ -278,6 +278,7 @@ const localizeInit = () => {
                     // Color panel
                     'colors': 'COLORS',
                     'colors.tint': 'Tint',
+                    'colors.temperature': 'Temperature',
```
willeastcott, line 281: What about the other localizations?

src/ui/color-panel.ts
```diff
@@ -74,6 +74,28 @@ class ColorPanel extends Container {
...
+
+        const temperatureLabel = new Label({
+            text: localize('colors.temperature'),
+            class: 'color-panel-row-label'
+        });
+
+        const temperatureSlider = new MyFancySliderInput({
+            class: 'color-panel-row-slider',
+            min: -0.3,
+            max: 0.3,
```
willeastcott, line 91: Is this a reasonable range?

src/ui/color-panel.ts
```diff
@@ -74,6 +74,28 @@ class ColorPanel extends Container {
...
+        const temperatureRow = new Container({
+            class: 'color-panel-row'
+        });
+
+        const temperatureLabel = new Label({
+            text: localize('colors.temperature'),
+            class: 'color-panel-row-label'
+        });
+
+        const temperatureSlider = new MyFancySliderInput({
```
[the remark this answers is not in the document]
slimbuck: That's my fault :)

src/ui/localization.ts
slimbuck: I never added the other for colors, will do it separately.

src/ui/color-panel.ts
slimbuck: Updated to -0.5...0.5

src/ui/localization.ts
simonbethke: I wasn't the one inventing the not-translated color-panel

src/ui/color-panel.ts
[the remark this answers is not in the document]
simonbethke: I thought the same when using that component

PR #478, opened 2025-04-17
Implementation of a render image dialog

simonbethke:
I created a simple image export dialog with the goal, to support larger resolutions than the viewport of my browser. By doing that, I found that I could/should include more settings that have been copied from the video-export code.

Screenshot 2025-04-22 at 16 01 13

### DIFF
--- a/src/editor.ts
+++ b/src/editor.ts
@@ -35,6 +35,10 @@ const registerEditorEvents = (events: Events, editHistory: EditHistory, scene: S
         return msg;
     });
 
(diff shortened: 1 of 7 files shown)


simonbethke:
The alignment of the dialog items is still bad, but I am not sure what your best practice is.
I am not sure if the old quick screenshot should still exist. The Render Menu could have two different render image items. Also the quick image could have a distinct shortcut.

willeastcott (approved):

PR #513, opened 2025-06-18
Update deps

slimbuck:
This PR updates all package dependencies, including the engine to latest.

The build-time ENGINE_PATH and PCUI_PATH options are removed and instead we rely on npm's link functionality.

A remaining issue is the following compile-time warning:

```
(!) [plugin typescript] @rollup/plugin-typescript TS5110: Option 'module' must be set to 'Node16' when option 'moduleResolution' is set to 'Node16'.
```

Making the suggested change results in other issues that I haven't managed to solve.


willeastcott (approved):
Great improvement.

PR #515, opened 2025-06-19
Type and lint fixes

kpal81xd:
- Fixed linting for `*.mjs` files
- Disabled type checking in rollup and split into separate type check script (run in CI now)


slimbuck (approved):

johnbanq:
Hi! I have two questions regarding this PR:

* In commit d7c5252952cc14a908dd7209d29e99cb2cdbc040, what is the reasoning for disabling rule 'import/no-unresolved' for all mjs files?
* commit 3445948fa8d4c2921531fab317e319ecb25eccc1 disabled type checking during rollup build and 45d314ff98f1845236c9264d54df6d7e9feef657 put that into CI instead. Is this done to reduce build time, or to fix the problem caused by PR #513 ?

Thanks in advance!

kpal81xd:
> Hi! I have two questions regarding this PR:
> 
> * In commit [d7c5252](https://github.com/playcanvas/supersplat/commit/d7c5252952cc14a908dd7209d29e99cb2cdbc040), what is the reasoning for disabling rule 'import/no-unresolved' for all mjs files?
> * commit [3445948](https://github.com/playcanvas/supersplat/commit/3445948fa8d4c2921531fab317e319ecb25eccc1) disabled type checking during rollup build and [45d314f](https://github.com/playcanvas/supersplat/commit/45d314ff98f1845236c9264d54df6d7e9feef657) put that into CI instead. Is this done to reduce build time, or to fix the problem caused by PR [Update deps #513](https://github.com/playcanvas/supersplat/pull/513) ?
> 
> Thanks in advance!

So the `import/no-resolved` was mainly for the `eslint.config.mjs`since the parsers import typing is invalid. It doesnt have to be scoped that wide I can just disable those particular import lines in the file

Technically its both - there was an import resolution issue with including external types for svg imports so disabled the typechecking in the rollup build and switched to the CI (related https://github.com/rollup/plugins/issues/1878) . Additionally this does speed up the build itself since type checking is done upfront

tsconfig.json
```diff
@@ -16,6 +16,7 @@
         "global.d.ts"
     ],
     "exclude": [
+        "src/debug.ts", // Testing file
```
johnbanq, line 18: Can we remove this? Since testing is done?

tsconfig.json
```diff
@@ -2,7 +2,7 @@
     "compilerOptions": {
         "target": "es2022",
         "module": "es2020",
-        "moduleResolution": "node16",
+        "moduleResolution": "bundler",
         "allowArbitraryExtensions": true,
```
johnbanq: I personally tried, it works without specifying "allowArbitraryExtensions": true, removing it also prevents type checker from registering import of unknown files as 'any', so we might want to remove it

johnbanq:
> > Hi! I have two questions regarding this PR:
> > 
> > * In commit [d7c5252](https://github.com/playcanvas/supersplat/commit/d7c5252952cc14a908dd7209d29e99cb2cdbc040), what is the reasoning for disabling rule 'import/no-unresolved' for all mjs files?
> > * commit [3445948](https://github.com/playcanvas/supersplat/commit/3445948fa8d4c2921531fab317e319ecb25eccc1) disabled type checking during rollup build and [45d314f](https://github.com/playcanvas/supersplat/commit/45d314ff98f1845236c9264d54df6d7e9feef657) put that into CI instead. Is this done to reduce build time, or to fix the problem caused by PR [Update deps #513](https://github.com/playcanvas/supersplat/pull/513) ?
> > 
> > Thanks in advance!
> 
> So the `import/no-resolved` was mainly for the `eslint.config.mjs`since the parsers import typing is invalid. It doesnt have to be scoped that wide I can just disable those particular import lines in the file
> 
> Technically its both - there was an import resolution issue with including external types for svg imports so disabled the typechecking in the rollup build and switched to the CI (related [rollup/plugins#1878](https://github.com/rollup/plugins/issues/1878)) . Additionally this does speed up the build itself since type checking is done upfront

Thanks for your detailed answer! I actually managed to get the type checking problem to go away this morning as well. I have left what I discovered as comments, hope this is helpful!

rollup.config.mjs
```diff
@@ -63,7 +64,12 @@ const application = {
             ]
         }),
         typescript({
-            tsconfig: './tsconfig.json'
+            tsconfig: './tsconfig.json',
+            compilerOptions: {
+                // FIXME: Disabled type checking as non ts file imports error due to global.d.ts not
+                // being imported.
+                noCheck: true
```
johnbanq: Pin @rollup/plugin-typescript to 12.1.2 will fix the global.d.ts issue as well, should we try that instead?

tsconfig.json
kpal81xd: Well the debug.ts file is still there so until thats removed ill leave it in

rollup.config.mjs
kpal81xd: Yep just tried and it works nicely thanks!

tsconfig.json
kpal81xd: Ahh okay yea i mainly added it for testing

tsconfig.json
kpal81xd: Acc Ill remove it since its skipped when bundling anyway now because it has no references

PR #520, opened 2025-06-22
Add orthographic projection option to image render

Evan-Dabeeo:
Add projection method to Image Rendering dialog.
Set cene.camera.ortho to true when "orthographic projection" is selected.


slimbuck:
Hi @Evan-Dabeeo ,

Thanks so much for this PR. I just tried your branch and it works well.

I wonder though if it would be better to just expose orthographic projection in the editor instead. This way user can create the screenshot precisely and not have to guess.

TBH when I added support for orthographic camera, I always assumed users would request to have general orthographic mode, but no one ever did. It should be very straightforward to add.

Evan-Dabeeo:
Hi. @slimbuck  Thanks for your feedback.

I think your suggestion is better. I think we should add a menu to select the projection method to View Option. I will check if it is possible.

PR #521, opened 2025-06-23
Add orthographic view button to right toolbar menu

Evan-Dabeeo:
Add orthographic view button to right toolbar menu

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -208,6 +208,11 @@ const main = async () => {
         scene.forceRender = true;
     });
 
+    events.on('camera.orthographicProjection', (projection: string) => {
+        scene.camera.ortho = true;
+        scene.forceRender = true;
+    });
+
     // initialize colors from application config
     const toColor = (value: { r: number, g: number, b: number, a: number }) => {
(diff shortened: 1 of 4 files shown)


Evan-Dabeeo:
At first, I wanted to add a UI to select the projection method in the view panel, but I found that when I selected orthographic projection and then moved the viewpoint, it switched back to perspective projection. Therefore, I felt the need to change it so that the orthographic projection view could only be seen once at that point, and I thought the optimal location for this would be the right toolbar. The icon doesn't quite fit, so I hope you can change it to something better.

slimbuck:
Hi @Evan-Dabeeo , 

I'm so sorry it's taken so long for me to check your PR.

I'm curious if you only want a ortho in this form (as in, temporarily till you rotate)? I would have expected a perspective/orthographic mode toggle would be more useful?

Thanks!

Evan-Dabeeo:
I think the function to temporarily switch to orthographic view is sufficient for now. A toggle view would be more useful, but since the existing view function switches to perspective view when the viewpoint is moved, if a toggle function is added, this part will also need to be considered and modified. I hope someone who needs this function will work on it later.

PR #555, opened 2025-08-08
Update deps

slimbuck:
Update dependencies


slimbuck:
Too soon, eslint failing now.

PR #578, opened 2025-09-17
Measurement tool Initial Implementations

nancywan1004:
My initial implementations for the measure tool. Core features include:
 - Activate the measure tool in the bottom toolbar and to see the distance input show up in the scene panel
 - Select a start and an end point to measure and see the distance in the scene panel
 - Change the value of the distance input to change the scale, and vice versa
 - Deactivate the tool and the line in the scene will disappear and the distance will be reset

To improve:
- The scale might tweak a minor bit as I make different measurements (i.e. sometimes 0.997, sometimes 1.011)
- Sometimes the line would disappear as I set the distance/scale directly (will need some investigations)
- The look of the line(how it renders) might not be the best for now

Would appreciate further feedback :)

Picture for reference:
截屏2025-09-17 14 09 13

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -23,6 +23,7 @@ import { RectSelection } from './tools/rect-selection';
 import { RotateTool } from './tools/rotate-tool';
 import { ScaleTool } from './tools/scale-tool';
 import { SphereSelection } from './tools/sphere-selection';
+import { MeasureTool } from './tools/measure-tool';
 import { ToolManager } from './tools/tool-manager';
 import { registerTransformHandlerEvents } from './transform-handler';
 import { EditorUI } from './ui/editor';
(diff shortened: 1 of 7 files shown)


AlexMorgand:
Nice contribution!

Maybe a minor request, some measurement tools like to add text above the measurement line.
I think it's useful when you add several measurements (like for a room).

Maybe future work.

slimbuck:
Thank you for this @nancywan1004! It's one of the highest requested features from our users.

I would like to make a few changes to your PR before merging. I will try get to this in the next few days.

The things I'd like to change:
- move the measure UI to the same place as the other tools, just above the bottom toolbar
- try improve the visuals
- possibly support 3 points
- possibly store the points per splat

slimbuck:
Hey @nancywan1004 , thanks again for this awesome PR. I finally got the time to implement a simple measurement tool ([here](#624)] using your PS as inspiration. Please give the new one a go and let me know if it needs more work!

PR #585, opened 2025-09-19
Remove the operation from the history when the splat is removed

mmjinglin163:
Remove the operation from the history when the splat is removed


slimbuck:
Thanks so much for this PR.

I will hopefully get a chance to test and merge it within the next few days!

slimbuck:
Hey @mmjinglin163 ,

I would like to suggest a few changes to this PR:
- register for `scene.elementRemoved` directly in EditHistory constructor
- add a function to`EditOp` interface for testing whether it applies to the supplied `Element` instance
- implement the function where necessary for `StateOp` and `MultiOp`

What do you think, does this sound reasonable?

If you could make these changes that would be great, otherwise I can too.

Thanks again!

mmjinglin163:
You're right—these changes will make the structure cleaner and reduce coupling.

PR #588, opened 2025-09-25
Speed up video recording

theamusing:
The video recording is slow when camera is not moving during some frames, because sorting does not work in this case, and the program will just wait until sorter time out
```
// in cases where the camera does not move between frames the sorter won't run
// and we need a timeout instead. this is a hack - the engine should allow us to
// know whether the sorter is running or not.
setTimeout(() => {
    resolve();
}, 1000);
```
which is annoying. So I recorded the camera position and forward for the last frame and skip sorting when their value don't change in the new frame.

### DIFF
--- a/src/render.ts
+++ b/src/render.ts
@@ -184,8 +184,8 @@ const registerRenderEvents = (scene: Scene, events: Events) => {
             const splats = (scene.getElementsByType(ElementType.splat) as Splat[]).filter(splat => splat.visible);
 
             // remember last camera position so we can skip sorting if the camera didn't move
-            let last_pos = new Vec3(0, 0, 0);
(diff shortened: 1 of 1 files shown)


theamusing:
Besides I'm working on adding timeline animations for splats (in case user just want get a video of splats rotating and don't want to adjust camera.) If you think this feature is needed, I'll try to pull a request later.

slimbuck:
This is great thanks so much!

slimbuck:
I will test and hopefully merge today

theamusing:
@slimbuck Thanks for your time and consideration!

slimbuck (approved):
Tested, works - thank you!

PR #590, opened 2025-09-27
Add timeline animation for entity

theamusing:
### New feature: Added animation for splat entity
Timeline keyframes can now record position, rotation and scales of entities and interpolate to get animations.
Also serialized animation data into document.

https://github.com/user-attachments/assets/97d4e7d7-81ea-4b09-933b-f4644f5a0bdf

Making it easier to create camera orbit animation.

### Bug Fix: drag keyframes in camera poses
The original version of drag keyframe has a bug: The pose list may be changed if removed first:
```
const movePose = (index: number, frame: number) => {
        // remove target frame pose
        const toIndex = poses.findIndex(p => p.frame === frame);
        if (toIndex !== -1) {
            removePose(toIndex);// List may be changed!!!
        }

        // move pose
        poses[index].frame = frame;
        rebuildSpline();
        events.fire('timeline.setKey', index, frame);
    };
```
So the pose on index should be assigned new value before removing.

### DIFF
--- a/src/camera-poses.ts
+++ b/src/camera-poses.ts
@@ -104,12 +104,12 @@ const registerCameraPosesEvents = (events: Events) => {
     const movePose = (index: number, frame: number) => {
         // remove target frame pose
         const toIndex = poses.findIndex(p => p.frame === frame);
+        // move pose
+        poses[index].frame = frame;
         if (toIndex !== -1) {
             removePose(toIndex);
         }
 
-        // move pose
-        poses[index].frame = frame;
         rebuildSpline();
         events.fire('timeline.setKey', index, frame);
     };
--- a/src/doc.ts
+++ b/src/doc.ts
@@ -118,8 +118,11 @@ const registerDocEvents = (scene: Scene, events: Events) => {
                 console.error('this should never fire');
             }
 
+            const splats = events.invoke('scene.allSplats') as Splat[];
+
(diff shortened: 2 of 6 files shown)


theamusing:
#591

slimbuck:
Hi @theamusing,

This is very interesting! I am wondering what the use-case is for animated models?

TBH I'd love to implement a full-on animation system, but I'm not really sure how many users needs that...

Is there any change you could submit the bugfix as a separate PR? 🙏🏻 I think that needs merging asap.

theamusing:
Hi @slimbuck,
Thanks for your interest. Here are the two main reasons I want model animation:

- I used Supersplat to display 3DGS models and compare their differences from various view directions by rotating them. With model animation, this can be easily done and rendered into a single video. Otherwise, I would need to render multiple videos and then merge them.

- I want to create a camera orbiting animation (in a circle). Using the original method of interpolating camera positions with a cubic spline requires many keyframes to approximate a circle. However, with scene animation (interpolating the quaternion), it can be done much more easily.

Really appreciate your time and consideration! Supersplat is such a great project and I'm willing to contribute to it.

theamusing:
#593 A seperate PR for bug fix. @slimbuck

connornishijima:
I could absolutely benefit from model animation ❤️ Thanks for your hard work

PR #593, opened 2025-09-29
Fix movePose bug

theamusing:
### Bug Fix: drag keyframes in camera poses
The original version of drag keyframe has a bug: The pose list may be changed if removed first:
```
const movePose = (index: number, frame: number) => {
        // remove target frame pose
        const toIndex = poses.findIndex(p => p.frame === frame);
        if (toIndex !== -1) {
            removePose(toIndex);// List may be changed!!!
        }

        // move pose
        poses[index].frame = frame;
        rebuildSpline();
        events.fire('timeline.setKey', index, frame);
    };
```
So the pose on index should be assigned new value before removing.


slimbuck:
THANK YOU!!

PR #595, opened 2025-10-02
Cursor hints for add and remove

simonbethke:
I created some svg cursors that feature a + and - sign to indicate if a modifier (Shift or CTRL) is pressed

https://github.com/user-attachments/assets/4115c5d2-ea2d-4e58-bfd8-9578f265f3f2


slimbuck:
thanks so much for this PR @simonbethke ! It was the basis for #957 which includes intersection (ctrl+shift) modifier.

PR #596, opened 2025-10-04
New Flood Select Tool

simonbethke:
I created a tool that allows to select splats with a flood-select. This works similar to the bucket-fill in painting apps. As the area is first only shown as masked and I can still set, add or remove it from the selection, there is also a threshold that can be adjusted. For easy usage, it is adjustable by using the mouse-wheel.

https://github.com/user-attachments/assets/9efbbd2a-828e-48fb-ad84-6139fd77c835


slimbuck:
This doesn't seem to work for me.

Am I doing something wrong?

https://github.com/user-attachments/assets/edd6082a-3bc4-42b6-b07d-db741421757c

@simonbethke I'm wondering whether we need set/add/remove buttons or whether relying on shift+ctrl would be better?

slimbuck:
Oh ignore that - I had a dodgy local engine.

simonbethke:
Okay, this is the first iteration. I created a second PR for the second iteration that is more userfriedly but requires more modification in pointer-capturing.

slimbuck:
Resolved in #598

PR #598, opened 2025-10-04
Flood-select-immediate-mode

simonbethke:
This adds an immediate mode to the flood select.
The normal flood select is fine for showing what is going on. By changing the threshold, the user gets an understanding what this tool is doing. 

However, for efficient editing, you want to select, delete, rotate, select, delete, rotate... and so on. With the normal flood select this is tideous, because every select needs a click at the bottom and to rotate the view you have to leave the tool.

With the immediate mode, the set/add/remove buttons are disabled. Instead of seeing the mask, the masked splats are immediately selected. Modifier keys shift and ctrl can be used to add/remove the splats from the current selection. Additionally, 3D-View navigation stays active.


simonbethke:
I have made this a separate PR because I think that moving the setPointerCapture() from the pointerdown to the first pointermove after pointerdown might be controversal.
The idea is, that this enables us to perform either rotate or a click with the left mouse-button without any modifier. 

It might also be controversal if/how this works with a touch-device.

slimbuck:
This version feels pretty natural to me compared to the previous one!

Few thoughts:
- I'd like the tool to just be immediate mode (and remove the set/add/remove and immediate toggle)
- pr includes extra files, svg.bak and viewer submodule
- flood selection code is doing some weird and unused RED/BLUE stuff. would be good to remove that.
- I'm not sure about the controller change, can we go without that for the first version?

simonbethke:
Okay, I cannot continue on this today, but here are some comments on your thoughts:

- Having the tool always in immediate mode is not that nice, because you get no idea what the threshold is for. Maybe if the threshold was a real slider (not this text-field) I could show the mask only during manipulation of that slider.
- Extrafiles are crap. The bak-file appears when editing something with notepad++ but I have no idea how to deal with that viewer submodule... hmmmm.... maybe I needed that to get it running at all? I dunno
- Unused RED/BLUE stuff? I dont know what you mean. I will check... if its obvious I will remove it :)
- The immediate mode doesnt work without the controller-change. It only makes sense if you can adjust the view between the clicks. But if you capture the pointer on mouse down for rotation of the view, you cannot listen to that same mouse-down on the editor-element for the flood select. I knew that this specific change is a risky one and has the potential to cause a ton of regressions. So far I didnt see any, but here be dragons.

slimbuck (approved):
Awesome thank you!

PR #599, opened 2025-10-07
Add preview feature on export popup

hisayan:
My idea is a preview method that uses scene.compressed.ply and setting.json via the cache.

[demo video](https://x.com/anotherbrain/status/1975361370150613229)


simonbethke:
I think this would be nice if the preview would run in an iframe in a supersplat-dialog-box. Maybe even extend that export dialog and integrate the preview in there

Issue #504, opened 2025-06-10
render difference between mac and windows


marlinilram:
Hi, we have noticed that the rendering results of PLY files differ between macOS and Windows, and this phenomenon occurs to varying degrees with different PLY files. 

![Image](https://github.com/user-attachments/assets/8e3d5ecf-5b19-42b9-b672-69330480e048)
![Image](https://github.com/user-attachments/assets/83404a77-1a44-45da-adef-048533fff1ea)

Any idea how this happens?
Thank you very much!

slimbuck:
Hi @marlinilram ,

On metal / macos, multiple layers of gaussians are blended together with high-precision even though we output to an RGBA8 buffer.

On windows though this doesn't happen. Blending is performed as you would expect on a RGBA8 buffer directly. With many layers of blending the error accumulates and you see discoloration.

You can fix this in your app by rendering to an RGBA16F or RGBA32F backbuffer instead of RGBA8.

However, this will result in more memory being allocated and slower rendering.

I have only seen this blending error on scenes containing a large number of semitransparent overlapping gaussians. I'm not sure why or when a training generates this. If possible, try to fix your training pipeline to output more solid/accurate gaussians. Then you can stick to RGBA8 without the discoloration.

Thanks

richardrevesz:
> Hi [@marlinilram](https://github.com/marlinilram) ,
> 
> On metal / macos, multiple layers of gaussians are blended together with high-precision even though we output to an RGBA8 buffer.
> 
> On windows though this doesn't happen. Blending is performed as you would expect on a RGBA8 buffer directly. With many layers of blending the error accumulates and you see discoloration.
> 
> You can fix this in your app by rendering to an RGBA16F or RGBA32F backbuffer instead of RGBA8.
> 
> However, this will result in more memory being allocated and slower rendering.
> 
> I have only seen this blending error on scenes containing a large number of semitransparent overlapping gaussians. I'm not sure why or when a training generates this. If possible, try to fix your training pipeline to output more solid/accurate gaussians. Then you can stick to RGBA8 without the discoloration.
> 
> Thanks

"You can fix this in your app by rendering to an RGBA16F or RGBA32F backbuffer instead of RGBA8."
Hi, how do I exactly do that? I am using the exported viewer from the web editor. What do I need to edit?

slimbuck:
Hi @richardrevesz, I'm really sorry this isn't currently easily possible. We'll investigate though and figure something out.

[...]

Issue #465, opened 2025-04-10
Windows rendering issue


slimbuck:
The same PLY sometimes looks different (incorrect) on windows Chrome compared to macos Chrome:

![Image](https://github.com/user-attachments/assets/a30f9d53-c83e-4ec2-9682-44cd89053e90)

This issue was reported by Henry Pearce.

simonbethke:
@slimbuck when looking at the [getContext()](https://github.com/search?q=org%3Aplaycanvas+getContext&type=code) methods, there is no color-space defined. I would suggest to add the color-space [as suggested by the LLM](https://grok.com/share/bGVnYWN5_8bbc40a7-24ed-41ce-b1be-cea6b5fe5eb7).

slimbuck:
It turns out that mac hardware actually accumulates/blends to a 16bit tiled memory buffer before writing to the 8-bit target.

The discoloration only appears on the NVIDIA hardware because the GPU is actually accumulating/blending to RGBA8, resulting in rounding off over multiple layers.

The discoloration seems to be made worse by our use of premultiplied alpha blending, but I currently don't really understand why (see #421). Presumably blending hardware on NVIDIA is low-precision.

willeastcott:
Fixed by #612. Closing.

PR #612, opened 2025-10-16
Support high precision render target

slimbuck:
Fixes: #465
Fixes: #504

Switch to using floating point render target instead of RGBA8. This is slower to render and consumes more memory, but helps address issues with color degradation in some scenes.

This PR:
- adds a toggle to render rgbaf16 instead of rgba8
- default to on
- handle video and image capture correctly

### DIFF
--- a/src/ui/view-panel.ts
+++ b/src/ui/view-panel.ts
@@ -234,7 +234,7 @@ class ViewPanel extends Container {
             class: 'view-panel-row-toggle',
             value: true
         });
-        
+
         highPrecisionRow.append(highPrecisionLabel);
         highPrecisionRow.append(highPrecisionToggle);
 


src/camera.ts
```diff
@@ -363,9 +364,10 @@ class Camera extends Element {
     rebuildRenderTargets() {
         const device = this.scene.graphicsDevice;
         const { width, height } = this.targetSize ?? this.scene.targetSize;
+        const format = this.scene.events.invoke('camera.highPrecision') ? PIXELFORMAT_RGBA16F : PIXELFORMAT_RGBA8;
 
         const rt = this.entity.camera.renderTarget;
-        if (rt && rt.width === width && rt.height === height) {
+        if (rt && rt.width === width && rt.height === height && rt.colorBuffer.format === format) {
```
mvaligursky, line 370: ideally use GraphicsDevice.getRenderableHdrFormat to make sure format is supported

mvaligursky: even though considering you don't use filtering, it would be supported on 99% of devices (apart from some iPhones)

PR #616, opened 2025-10-17
Stream video to disk when possible

slimbuck:


src/ui/editor.ts
```diff
@@ -236,7 +240,35 @@ class EditorUI {
...
+                        types: [{
+                            description: 'MP4 Video',
+                            accept: {
+                                'video/mp4': ['.mp4']
+                            }
+                        }],
+                        suggestedName: suggested
+                    });
+
+                    await events.invoke('render.video', videoSettings, fileHandle && await fileHandle.createWritable());
```
willeastcott: Maybe it's just me, but I'd find something like this more readable:

```
// Prompt the user to select a file for saving the MP4 video.
// If the File System Access API is unavailable, skip this step.
let writable = null;

if (window.showSaveFilePicker) {
    const fileHandle = await window.showSaveFilePicker({
        id: 'SuperSplatVideoFileExport',
        types: [{
            description: 'MP4 Video',
            accept: { 'video/mp4': ['.mp4'] }
        }],
        suggestedName: suggested
    });

    writable = await fileHandle.createWritable();
}

// Pass the writable stream (or null) to the render system.
await events.invoke('render.video', videoSettings, writable);
```

slimbuck: done

PR #633, opened 2025-10-28
Improve LCC loading

slimbuck:
This PR:
- updates the LCC loader to stream in the required parts of the data instead of the entire file
- adds a work-in-progress read source system with streaming and arraybuffer capability

This fixes an issue where any LCC model with a large amount of spherical harmonic data would fail to load due to javascript and browser buffer size limits.

### DIFF
--- a/src/loaders/asset-source.ts
+++ b/src/loaders/asset-source.ts
@@ -17,7 +17,7 @@ const createReadSource = async (assetSource: AssetSource, start?: number, end?: 
     } else if (assetSource.contents) {
         source = assetSource.contents.slice(start, end);
     } else {
-        source = await fetch(assetSource.url, { headers: { 'Range': `bytes=${start}-${end - 1}` } });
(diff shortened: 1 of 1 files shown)


src/loaders/lcc.ts
```diff
@@ -345,15 +358,16 @@ const deserializeFromLcc = (param:LccParam) => {
 
 const loadLcc = async (assetSource: AssetSource) => {
     // .lcc
-    const text = await fetchText(assetSource);
+    const textSource = await createReadSource(assetSource);
+    const text = new TextDecoder().decode(await textSource.arrayBuffer());
     const meta = JSON.parse(text);
 
-    const isHasSH: boolean = meta.fileType === 'Quality' || !!(assetSource.mapFile('shcoef.bin'));
+    const isHasSH: boolean = meta.fileType === 'Quality' && !!(assetSource.mapFile('shcoef.bin'));
```
xgrids-devteam: Due to historical versioning reasons, `meta.fileType` may be undefined. We recommend using the following code:
const isHasSH: boolean =  !!(assetSource.mapFile('shcoef.bin'));
image

PR #634, opened 2025-11-03
Fix transform panel on ssproj load

naga-k:
After loading all the splats and deserializing their saved transforms, this updated code manually refreshes the pivot to match the loaded position. Without this, the pivot would still be at the default [0,0,0] position from when the splat was first auto-selected (before deserialize ran), and the Transform Panel would show incorrect values.


naga-k:
@slimbuck I see the CICD linter failed on main. Created a new PR #639 since this is already merged :)

PR #637, opened 2025-11-05
Fix colmap images.txt loader

naga-k:
Removed the`.filter((_, i) => i % 2 === 0) // only need the first line of each pair` line in loadImagesTxt of filehandler.

### DIFF
--- a/src/file-handler.ts
+++ b/src/file-handler.ts
@@ -187,34 +187,15 @@ const loadImagesTxt = async (file: ImportFile, events: Events) => {
     const response = new Response(file.contents);
     const text = await response.text();
 
-    const lines = text.split('\n');
-    const poses: {
-        w: number;
-        x: number;
-        y: number;
-        z: number;
-        tx: number;
-        ty: number;
-        tz: number;
-        name: string;
-        order: number;
-    }[] = [];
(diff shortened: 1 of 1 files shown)


slimbuck:
Hi @naga-k,

Thanks for submitting this fix, it's always appreciated.

Unfortunately:
- most of the changes are rewrites of the existing code, but worse, for no reason (looking at you llm)
- I think the fix is wrong anyway. I have an `images.txt` with hundreds of points in the points2d array so testing `if (parts.length < 10)` will not skip these lines.

The correct fix is to simply remove line 194 of the current file:
```
.filter((_, i) => i % 2 === 0) // only need the first line of each pair
````

If you have an images.txt that I can use for extra testing that'd be great and if you like to update this PR with the above suggestion that'd be great too.

Thanks!

naga-k:
Thanks, @slimbuck. I knew the 194 line was problematic but thought the other changes helped too. Should have investigated it further. I dont have any images.txts on the macbook im using today. will test it out soon.

I deleted the other changes and pushed a commit.

willeastcott:
Please update the original PR description and title to match the change.

naga-k:
@willeastcott updated PR title and description - done

slimbuck (approved):

PR #642, opened 2025-11-05
fix: Remove outdated <meta> tags

abstrakt8:
## What's changed?

Currently the OpenGraph and Twitter `<meta>` tags are outdated (e.g. wrong URL). 

There are also some other SEO relevant tags that I want to add, but are not relevant for this repository:
* Canonical `<link rel="canonical">`
* Structured JSON (e.g. breadcrumbs)

~~I added a placeholder line (`<!-- ssr-head -->`) that allows us to dynamically inject the SEO tags on the server side by using something like `html.replace('<!-- ssr-head -->', '...')`. This allows us to quickly make changes to SEO tags without having to touch this repository.~~

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -10,8 +10,6 @@
         <meta name="description" content="SuperSplat is an advanced browser-based editor for manipulating and optimizing 3D Gaussian Splats. It is open source and engine agnostic." />
         <meta name="viewport" content="width=device-width, user-scalable=no, minimum-scale=1.0, maximum-scale=1.0" />
 
-        <!-- ssr-head -->
-
         <!-- jszip -->
         <script src="jszip.js"></script>
 


src/index.html
```diff
@@ -8,22 +8,9 @@
...
-        <meta property="og:description" content="SuperSplat is an advanced browser-based editor for manipulating and optimizing 3D Gaussian Splats. It is open source and engine agnostic." />
-        <meta property="og:image" content="https://playcanvas.com/supersplat/editor/static/images/header.webp" />
-
-        <!-- Twitter -->
-        <meta property="twitter:card" content="summary_large_image" />
-        <meta property="twitter:url" content="https://playcanvas.com/supersplat/editor" />
-        <meta property="twitter:title" content="SuperSplat" />
-        <meta property="twitter:description" content="SuperSplat is an advanced browser-based editor for manipulating and optimizing 3D Gaussian Splats. It is open source and engine agnostic." />
-        <meta property="twitter:image" content="https://playcanvas.com/supersplat/editor/static/images/header.webp" />
+        <!-- ssr-head -->
```
willeastcott: I suspect you can inject meta tags in the head element without needing a string to match. You have the `</head>` for example.

abstrakt8: One little concern I had is that the order might be important. If I place at the end, it would be after the `<script>` tags and that might not be SEO friendly. 

I'll still place it at the end of <head> anyways and we'll see how it behaves.

slimbuck (approved):

src/index.html
willeastcott: Maybe inject after the opening `head` element then? 😄

PR #677, opened 2025-11-23
Externalize and dynamically load translations

willeastcott:
This update moves all text translations out of the main application code and into separate JSON files. Only the language needed is loaded now.


slimbuck:
I'm wondering what the motivation for this change is: better maintenance of the translations or load time performance or ...?

willeastcott:
Yeah, I should have also mentioned...the original localization.ts was getting very long indeed.

PR #697, opened 2025-11-29
Add eyedropper selection tool 

RELNO:
Implement an eyedropper selection tool that allows users to select splats based on color similarity, with a simple configurable color threshold for sensitivity.

Use case: Simple way to select "skies" when cleaning up a scene 

To do: 
- Improve threshold UI, radial/HSV based distance selector?
- Localize 

selecting whites in scene

### DIFF
--- a/static/locales/en.json
+++ b/static/locales/en.json
@@ -207,6 +207,7 @@
     "tooltip.bottom-toolbar.lasso": "Lasso Select",
     "tooltip.bottom-toolbar.polygon": "Polygon Select ( P )",
     "tooltip.bottom-toolbar.brush": "Brush Select ( B )",
+    "tooltip.bottom-toolbar.flood": "Flood Select ( O )",
     "tooltip.bottom-toolbar.eyedropper": "Eyedropper Select ( E )",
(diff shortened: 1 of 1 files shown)


slimbuck:
Sorry it's taken me so long to look at your PR.

I'm a little surprised by how well it works TBH. I am a little concerned about using this tool with models that have spherical harmonics as users might get some pretty unexpected results. However for many cases I think this is ace.

Thanks so much for the PR!

RELNO:
Thanks @slimbuck - code base is so well organized that it was quite simple to add it. 

The SH point is valid, but in some extra tests I did with the "material" [scene](https://huggingface.co/datasets/ZehaoYu/mip-splatting/blob/main/materials.ply) -- which has pretty distinct reflections and speculars -- I still saw pretty consistent selection results:
Screenshot 2025-12-02 at 10 50 10

This functionality is practically a way to replace the SPLAT DATA - RGB selection mode, with a more intuitive interface

slimbuck (approved):

slimbuck:
Amazing - thanks so much for this!

Btw I imagine we might want to add separate thresholds for HSV at some point.

PR #742, opened 2026-01-09
Update default settings

slimbuck:
## Update default settings

Updates several default values to improve the initial user experience:

- Change default camera FOV from 65° to 75°
- Set splat overlay to off by default
- Reduce default fly speed from 2 to 1

### Files Changed

- `src/scene-config.ts` - FOV and overlay defaults
- `src/camera.ts` - Fly speed class property
- `src/ui/view-panel.ts` - Fly speed slider initial value

### DIFF
--- a/src/scene-config.ts
+++ b/src/scene-config.ts
@@ -14,11 +14,11 @@ const sceneConfig = {
     camera: {
         pixelScale: 1,
         multisample: false,
-        fov: 65,
+        fov: 75,
         exposure: 1.0,
         toneMapping: 'linear',
         debugRender: '',
-        overlay: true,
+        overlay: false,
         highPrecision: true
     },
     show: {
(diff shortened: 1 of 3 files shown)


src/scene-config.ts
```diff
@@ -14,11 +14,11 @@ const sceneConfig = {
     camera: {
         pixelScale: 1,
         multisample: false,
-        fov: 65,
+        fov: 75,
```
abstrakt8, line 17: Do we need to change it here as well? 

https://github.com/playcanvas/supersplat-viewer/blob/main/src/settings.ts#L94

slimbuck: yeah could do

slimbuck: https://github.com/playcanvas/supersplat-viewer/pull/129

PR #753, opened 2026-01-15
Refactor keyboard shortcuts: use Ctrl for common shortcuts, Alt for camera slow

slimbuck:
# Refactor keyboard shortcuts: use Ctrl for common shortcuts, Alt for camera slow

## Summary

Changes several keyboard shortcuts from Alt-based to Ctrl-based modifiers for better cross-platform compatibility (Ctrl+A, Ctrl+D don't work on Windows when reserved by the system). The camera fly slow modifier is changed from Ctrl to Alt to free up Ctrl for these shortcuts.

### Shortcut Key Bindings

Updated shortcuts to use Ctrl instead of Alt:

| Action | Before | After |
|--------|--------|-------|
| Select All | Alt+A | Ctrl+A |
| Select None | Alt+Shift+A | Ctrl+Shift+A |
| Eyedropper Tool | Alt+E | Ctrl+E |
| Toggle Data Panel | Alt+D | Ctrl+D |

### Camera Fly Speed Modifiers

Changed the slow camera modifier from Ctrl to Alt:

| Modifier | Key | Effect |
|----------|-----|--------|
| Fast | Shift | 10x speed |
| Slow | Alt | 0.1x speed |

Renamed shortcut event IDs to semantic names:
- `camera.modifier.shift` → `camera.modifier.fast`
- `camera.modifier.ctrl` → `camera.modifier.slow`

### Input Handling Fixes

- Simplified focus check: shortcuts now always skip when focus is not on `document.body`
- Added `e.repeat` check to ignore repeated keydown events for non-held shortcuts, preventing conflicts with held keys like WASD when adding modifiers

## Files Changed

- `src/shortcut-manager.ts` - Updated default shortcut bindings
- `src/controllers.ts` - Renamed modifier event handlers and variables
- `src/shortcuts.ts` - Fixed input handling and repeat key logic
- `src/ui/shortcuts-popup.ts` - Updated camera hint display

### DIFF
--- a/src/controllers.ts
+++ b/src/controllers.ts
@@ -273,8 +273,8 @@ class PointerController {
         let flyUp = false;
 
         // track modifier keys for speed control (updated via shortcut events)
-        let shiftDown = false;
-        let ctrlDown = false;
+        let fastDown = false;
+        let slowDown = false;
 
         // Clear all keys when window loses focus to prevent stuck keys
         const clearAllKeys = () => {
@@ -284,8 +284,8 @@ class PointerController {
             flyRight = false;
             flyDown = false;
             flyUp = false;
-            shiftDown = false;
-            ctrlDown = false;
+            fastDown = false;
+            slowDown = false;
         };
 
         // Helper to switch to fly mode when a fly key is pressed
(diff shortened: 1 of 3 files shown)


LeXXik:
I'm late here, but such changes would be nice to test on Windows as well. For example, Ctrl+D is a bookmark shortcut in Windows Chrome.

slimbuck:
Please do check it.

I went to some trouble to get the key mappings to work on all platforms, browsers and also for non-qwerty keyboards, but may have missed something.

PR #754, opened 2026-01-16
Allow holding key to change brush select size

LeXXik:
Similar to Blender, allow holding the key to change the brush size, as mouse input doesn't always have a scroll.

- Waits a bit before the size starts to change, so user may step tap as previously.
- Frame time dependent, so higher refresh rates will make the size change faster. Didn't want to complicate the PR, but can be addressed in another one, if it becomes an issue.

https://github.com/user-attachments/assets/2683904d-8427-4ef3-b93b-6a65ab1316f4

### DIFF
--- a/src/tools/brush-selection.ts
+++ b/src/tools/brush-selection.ts
@@ -181,7 +181,7 @@ class BrushSelection {
             parent.removeEventListener('wheel', wheel);
         };
 
-        events.on('tool.brushSelection.smaller', (down: boolean) => {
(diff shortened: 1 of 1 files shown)


slimbuck:
ahh, this was actually recently broken due to [this line](https://github.com/playcanvas/supersplat/blob/main/src/shortcuts.ts#L86).

I think it'd probably be better to add another key option which is "allow repeat".

This would make it work the way it did before.

LeXXik:
Ah, so its a bug? Nevermind then. Just noticed it when chacking the Windows hotkeys. Closing it.

PR #798, opened 2026-02-11
Show the number of splats in the scene manager per object

simonbethke:


LeXXik:
package-lock was probably committed by accident

slimbuck:
It might be worth shortening the number with K/M because the label could end up taking lots of space.

PR #828, opened 2026-03-09
Add particle dissolve effect with MediaPipe hand gesture control

MoiraCodes:
- Dissolve shader effect that scatters gaussians in random directions
- MediaPipe HandLandmarker pinch-to-dissolve gesture control via webcam
- Particle panel UI with dissolve/reset buttons
- Toolbar buttons for particle panel and gesture toggle

### DIFF
--- a/src/gesture-controller.ts
+++ b/src/gesture-controller.ts
@@ -1,0 +1,202 @@
+import { Events } from './events';
+
+const THUMB_TIP = 4;
+const INDEX_TIP = 8;
+
+class GestureController {
+    private events: Events;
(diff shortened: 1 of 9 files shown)


MoiraCodes:
added mediapipe

willeastcott:
This sounds like it would be better as a PlayCanvas Engine example.

willeastcott:
As previously mentioned, this isn't a good fit for the SuperSplat Editor. If you'd like to do a 3DGS MediaPipe example in the Engine repo, go right ahead. Closing.

PR #881, opened 2026-05-04
Video/Image Renders - use splat file name as render file name

azadbal:
## Summary

Use the first imported splat's filename as the suggested save name for rendered images and videos.

## What changed

- derive render save names from the first scene splat instead of the document name or `supersplat`
- preserve the existing `showSaveFilePicker` flow so saving to macOS locations like Downloads and Documents continues to work
- keep the existing browser download fallback for environments without the file save picker
- share the render filename logic between image and video export paths

## Notes

This PR intentionally keeps the save flow close to stock `supersplat`.
It updates the suggested filename, but does not add folder inspection or auto-incremented `(1)`, `(2)` collision handling.

## Validation

- `npm run lint`
- `npm run build`
- browser sanity check in Chrome: importing `splat1.ply` suggests `splat1.png` in the save picker
- manual verification that the same branch no longer hits the folder-access error caused by `showDirectoryPicker`

### DIFF
--- a/src/render-filename.ts
+++ b/src/render-filename.ts
@@ -1,0 +1,45 @@
+import { path } from 'playcanvas';
+
+import type { Splat } from './splat';
+
+const removeExtension = (filename: string) => {
+    return filename.substring(0, filename.length - path.getExtension(filename).length);
+};
+
+const isInvalidFilenameChar = (char: string) => {
+    return /[<>:"/\\|?*]/.test(char) || char.charCodeAt(0) < 32;
+};
+
+const sanitizeFilename = (filename: string) => {
+    const sanitized = Array.from(filename, (char) => {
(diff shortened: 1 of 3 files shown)


willeastcott:
Thanks @azadbal! This shipped in #959 with a slightly different naming rule — the project document name still takes priority when set, and the splat's (renamable) display name is used rather than the source filename — but your `sanitizeFilename` and URL-basename handling were ported nearly verbatim, with co-author credit on the commit. Closing since main has moved substantially underneath this branch (PNG→WebP render output, localization changes, 360 capture). The name-hygiene helpers were a genuinely good catch — much appreciated.

azadbal:
Happy to see it merged and glad it was helpful!

PR #885, opened 2026-05-08
Drive splat orientation from DataTable.transform instead of per-format hardcoding

slimbuck:
## Summary
- Replace the hardcoded per-format `getOrientation(filename)` switch in `asset-loader.ts` with the `transform` returned by splat-transform's readers, so each format's authoring frame is described by data rather than baked into the editor.
- `loadGSplatData` now returns `{ gsplatData, transform }`; the asset loader hands the `Quat` rotation straight to `Splat`.
- `Splat`'s constructor takes a `Quat` and applies it via `setLocalRotation`, replacing the previous `Vec3` Euler + `setEulerAngles` (avoids the deg→quat→deg round-trip and removes Euler ambiguity).
## Why
The `getOrientation` switch returned `(0, 0, 180)` for PLY/splat/KSplat/SOG, `(90, 0, 180)` for LCC, and `(0, 0, 0)` for SPZ. Splat-transform already tags every DataTable with the correct authoring-frame `Transform` (PLY/splat/KSplat/SOG/SPZ → `Transform.PLY = (0, 0, 180)`; LCC → `(90, 0, 180)`), so the per-format switch was duplicating information that the loader already provides — and the SPZ entry was wrong (this is what made SPZ files load 180° rotated relative to the equivalent PLY).
## Compatibility
Requires `@playcanvas/splat-transform` with the SPZ coordinate-system fix (in flight at https://github.com/playcanvas/splat-transform/pull/230). With that version, SPZ files load with the same `(0, 0, 180)` orientation as PLY and the two render identically.
## Test plan
- [ ] Load a `.ply` scene — orientation matches previous behavior.
- [ ] Load the same scene as `.spz` — now orients identically to the PLY (was 180° around Y before).
- [ ] Load a `.lcc` scene — orientation matches previous behavior (90, 0, 180).
- [ ] Load `.sog`, `.compressed.ply`, `.ksplat`, `.splat` — all orient as before.


josetiburcio55-cpu (commented):
D

PR #887, opened 2026-05-09
fix: normalize dblclick coordinates for Safari cross-browser compatibility

akshadjaiswal:
## Summary

Safari fires `dblclick` with the `canvas` element as `event.target` instead of the overlay element (`#canvas-container`). This means `offsetX`/`offsetY` are relative to the canvas, not the overlay — so `pickFocalPoint` receives wrong normalized coordinates on Safari, causing the focal point to land at the incorrect position when double-clicking to focus.

## What changed

- `src/controllers.ts`: detect when `event.target === canvas` (Safari path) and compute `x/y` from `clientX/clientY - target.getBoundingClientRect()` instead of using `offsetX/offsetY` directly
- Chrome path unchanged — `offsetX/offsetY` used as before since `event.target === target` there
- Replaces the `FIXME` comment with an explanation of why the two browsers differ

## Validation

- `npm run lint` — clean
- `npm run build` — clean
- Chrome: double-click to focus behaviour unchanged (Chrome path not touched)
- Safari: focal point now lands at the correct position matching the clicked point

### DIFF
--- a/src/controllers.ts
+++ b/src/controllers.ts
@@ -299,7 +299,9 @@ class PointerController {
             event.preventDefault();
         };
 
-        // FIXME: safari sends canvas as target of dblclick event but chrome sends the target element
+        // Safari fires dblclick with the canvas element as event.target instead of the
+        // overlay element (target). offsetX/offsetY are then relative to the canvas, not
+        // target, so we must normalize to target-space via clientX/clientY.
         const canvas = camera.scene.app.graphicsDevice.canvas;
 
         const dblclick = (event: globalThis.MouseEvent) => {
@@ -308,7 +310,16 @@ class PointerController {
                 if (camera.controlMode === 'fly') {
                     camera.scene.events.fire('camera.setControlMode', 'orbit');
                 }
-                camera.pickFocalPoint(event.offsetX / target.clientWidth, event.offsetY / target.clientHeight);
+
+                let x = event.offsetX;
+                let y = event.offsetY;
+                if (event.target === canvas) {
+                    const tr = target.getBoundingClientRect();
+                    x = event.clientX - tr.left;
+                    y = event.clientY - tr.top;
+                }
+
+                camera.pickFocalPoint(x / target.clientWidth, y / target.clientHeight);
             }
         };
 


akshadjaiswal:
Hi @slimbuck, just checking in on this — happy to make any changes if needed.

slimbuck:
Hi @akshadjaiswal ,

Sorry I'm not able to reproduce this issue.

Can you give me the steps?

Thanks!

akshadjaiswal:
Hi @slimbuck, thanks for looking at this!

The bug is Safari-specific, so it won't reproduce in Chrome/Firefox. Here are the steps:

**To reproduce the original bug (before this fix):**
1. Open SuperSplat in Safari
2. Load any `.ply` splat file
3. Double-click anywhere on the canvas to set the focal point
4. The focal point lands at the wrong position — offset from where you actually clicked

**Why it happens:**
Safari fires `dblclick` with the `<canvas>` element as `event.target`, while Chrome fires it with the overlay element (`#canvas-container`). When `canvas` is the target, `offsetX`/`offsetY` are relative to the canvas bounds — not the overlay. Since the fix was already in place to *check* for both targets in the condition (`event.target === target || event.target === canvas`), the handler fires correctly on Safari, but the coordinates passed to `pickFocalPoint` are wrong.

**To verify the fix:**
- On Safari, double-click after applying the fix — focal point now lands exactly where you clicked
- On Chrome, behaviour is completely unchanged (Chrome path not touched)

If you don't have Safari handy, the coordinate math can be verified logically: `clientX - getBoundingClientRect().left` is exactly how a browser computes `offsetX` when the target is that element — so both paths produce identical normalized values for the same physical click.

Happy to add a code comment clarifying this further if that helps!

slimbuck:
Yes I tried this on my mac m4. It works fine.

akshadjaiswal:
Thanks for testing! Just to clarify — did the test include Safari specifically? The existing condition already handles both targets correctly for the *handler to fire*, but the bug is specifically that `offsetX`/`offsetY` values are wrong when Safari sets `canvas` as the target. If it works correctly on Safari too, it's possible the canvas and overlay are perfectly aligned on M4 Macs (same bounding rect), which would make the coordinate offset zero and the bug invisible. Happy to close this if you're satisfied it's not an issue worth fixing.

slimbuck:
I'm going to close this issue under the assumption you haven't reproduced the issue at all and it's the result of an LLM.

PR #904, opened 2026-05-17
fix: volume select ignores splats outside the camera frustum

### LINKED ISSUE #843: Issue: Volume select should ignore camera view
When using Sphere or Box selection, the selection is only correct if the complete volume of the selection is visible in the editor before clicking the add button.
If you position the selection volume in the scene, look away from it and then click the add button, nothing gets selected.
If the selection volume is only partially visible before clicking the add button, only the visible parts of the selection volume influence which splats get selected.
Expected behaviour would be that all splats inside the volume get selected, no matter what the editor camera is currently looking at.

mvanhorn:
## Summary

Restructure `src/shaders/intersection-shader.ts` so the NDC frustum-cull check only gates the screen-space selection modes. Sphere (mode 2) and box (mode 3) now evaluate their world-space predicate unconditionally, so splats inside a volume are selected even when the camera is pointed away from them.

## Why this matters

Issue #843 reports that Sphere and Box volume selection only catches splats whose projection sits inside the current camera view - rotate the camera away from the volume and clicking Add selects nothing. Maintainer @slimbuck confirmed on 2026-03-18: "omg you're right. i can't believe it." The bug is the unconditional `if (!any(greaterThan(abs(ndc), vec3(1.0))))` frustum cull that wraps every mode-specific predicate. That gate is correct for rect / mask / pixel-precise selection (modes 0 and 1) which operate in screen space, but it is wrong for sphere and box which test world-space containment.

## Implementation notes

- World position is still hoisted before the mode switch so all four modes can read it.
- NDC computation plus the `!any(greaterThan(abs(ndc), vec3(1.0)))` cull moved inside an `(mode == 0 || mode == 1)` branch.
- Sphere (mode 2) and box (mode 3) branches now run their world-space tests without the NDC gate.
- Shader uniforms, output channel layout, and the CPU dispatch path in `src/data-processor/intersect.ts` and `src/editor.ts` are unchanged.

## Testing

- `npm run lint` clean.
- `npx tsc --noEmit` clean.
- Walked the diff against the scenarios from the issue:
  - Sphere / box volume placed behind the camera: world-space predicate runs without the NDC gate, splats inside the volume select.
  - Rect / lasso / brush (modes 0 and 1): NDC gate still in place, off-screen splats are still rejected.
  - `select.byMask` (mode 0): mask sampling still requires valid NDC; behavior unchanged.

Fixes #843

### DIFF
--- a/src/shaders/intersection-shader.ts
+++ b/src/shaders/intersection-shader.ts
@@ -71,38 +71,41 @@ const fragmentShader = /* glsl */ `
                 center = vec4(center, 1.0) * t;
             }
 
-            // transform to clip space and discard if outside
+            // transform to world space (sphere/box modes test world-space containment)
             vec3 world = (matrix_model * vec4(center, 1.0)).xyz;
-            vec4 clip = matrix_viewProjection * vec4(world, 1.0);
-            vec3 ndc = clip.xyz / clip.w;
 
-            // skip offscreen fragments
-            if (!any(greaterThan(abs(ndc), vec3(1.0)))) {
-                if (mode == 0) {
-                    // select by mask
-                    ivec2 maskUV = ivec2((ndc.xy * vec2(0.5, -0.5) + 0.5) * mask_params);
-                    clr[i] = texelFetch(mask, maskUV, 0).a < 1.0 ? 0.0 : 1.0;
-                } else if (mode == 1) {
-                    // select by rect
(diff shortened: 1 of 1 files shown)


willeastcott (approved):

willeastcott:
Thank you very much for the contribution, @mvanhorn!

mvanhorn:
Thanks for taking this, @willeastcott - volume select respecting the frustum makes big scenes much saner.

willeastcott:
No problem, @mvanhorn! 🙌

PR #933, opened 2026-07-02
Add 360 equirectangular video rendering

willeastcott:
## What

https://www.youtube.com/watch?v=6oGjT8eubMc

`Render > Video` gains a **Projection** option. Selecting **360° Equirectangular** exports a monoscopic equirectangular video (2:1 presets: 2048x1024, 3840x1920, 4096x2048) that plays on YouTube 360, VR headsets and VLC. A **Level Horizon** checkbox (default on) keeps the output level using only camera yaw; unchecked, full camera orientation (yaw + pitch) is baked into the projection.

## How

- **Capture** (`equirect-renderer.ts`, `shaders/equirect-shader.ts`): per output frame the scene is rendered into six 90° cube faces from the animated camera position, with splats re-sorted per face direction (sorting is projected-depth based, so each view direction needs its own sort). A fullscreen shader with manual face selection projects the six 2D face textures to the equirect frame, which feeds the existing WebCodecs/mediabunny pipeline unchanged. Using six 2D textures instead of a GL cubemap sidesteps cubemap face-orientation/flipY conventions entirely.
- **Camera** (`camera.ts`): new `poseOverride` applied inside `onUpdate`, since face orientations (arbitrary rotations incl. roll) cannot be expressed by the azim/elev pose system and direct entity writes are clobbered every tick. All six faces share identical direction-independent clipping planes, preventing culling seams at face boundaries. A new `displayTransform` holds the tween-driven pose while an override is active; the scene `prerender` event now broadcasts it so the view cube and overlays do not flicker through capture poses during renders.
- **Metadata** (`spherical-metadata.ts`): 360 MP4/MOV exports are tagged with Spherical Video V1 metadata (Google spatial-media spec) so players auto-detect the projection. These exports render to an in-memory buffer with `fastStart: false` (moov written last), then a `uuid` box is spliced into the video trak and ancestor box sizes patched. On any parse failure the untagged buffer is written instead, so a render is never lost to a tagging bug. WebM/MKV export untagged via the existing streaming path.
- **Dialog**: the Projection select swaps the resolution presets (2:1, capped at 4096 wide for H.264 L5.1 / hardware encoder limits), hides Portrait Mode and Show Debug Overlays, and shows Level Horizon. New strings localized across all 9 locales. Both render paths now pre-check `VideoEncoder.isConfigSupported` to fail fast with the standard error popup.

## Verified

- Orientation validated by rendering a scene and decoding the MP4 back to pixels: a forward object lands at image center, an object 90° to the right lands at u=0.75 (not mirrored), nadir/zenith map to the poles (validating both ±Y face bases), and objects straddling a face boundary render without tearing.
- Level Horizon: with a 44°-pitched camera, ON keeps the horizon level; OFF re-centers the view on the camera forward.
- The tagged MP4 demuxes and decodes in Chromium (proving moov size patching); the `uuid` box is byte-exact per the spherical video spec inside `moov > trak`.
- Cancellation restores fov/ortho/pose override and leaves no file; the standard render path is regression-tested; lint and production build pass.

## Notes for reviewers

- 360 renders are ~6x slower than standard (6 sorts + 6 locked render frames per output frame); progress reports fractionally per face and cancellation is checked per face.
- Tagged exports buffer the full MP4 in memory before writing (tens of MB for typical clips).
- Known limitation: splats intersecting the camera position can pop between faces (inherent to direction-sorted gaussians; radial sorting in the engine would fix this seam class for good).
- Future work: stereo ODS, >4K presets for HEVC/AV1, 360 image export via the same capture class.


slimbuck (approved):

willeastcott:
Follow-up commit: face-boundary seams were visible in real scenes (splats rasterize with slightly different projected shapes per face, and composite in a different sort order per face). Faces are now rendered at 100 degree fov and the equirect projection blends the overlap, feathering weights from 40 degrees off-axis to zero at the face edge. Weight coverage verified (min total weight 1.0 over the sphere, exact 50/50 at boundaries, thirds at cube corners) and the orientation test suite is unchanged.

grafpez:
this is great, can't wait to try this; H.265 / 8K+ resolutions will be critical though since 4K max at 360 isn't a satisfying pixel-per-degree density

willeastcott:
> this is great, can't wait to try this; H.265 / 8K+ resolutions will be critical though since 4K max at 360 isn't a satisfying pixel-per-degree density

I guess I could try adding 8k. Just a bit concerned about processing time. My test rendered a 3 minute 4k video at 60fps in about 100 minutes. So presumably this would at least quadruple that.

grafpez:
> I guess I could try adding 8k. Just a bit concerned about processing time. My test rendered a 3 minute 4k video at 60fps in about 100 minutes. So presumably this would at least quadruple that.
With 8K being 4x the data this can be expected. I would even go as far as suggesting a 90fps option would be great (to deliver to AVP) I guess the people who need/want this are already acutely aware of this being demanding on system resources (last 3DGS processing I did took roughly 10hrs..)

PR #938, opened 2026-07-04
Add camera info overlay showing live camera position and target

### LINKED ISSUE #605: Camera Information
With the new progression with splat-transform:  it is important to know where the camera is located for a particular view.  

Proposal is for a panel displaying this information.  Would also be worth having a 'copy' facility to put the information into the copy/paste buffer in the correct format for use.  The sample I have made in my own local copy works well and is toggled on/off using the I  ( as in info) key  - Right-Click on the line of info to copy to buffer

Image

willeastcott:
Fixes #605

camera-info-overlay

### What

Adds a **camera info overlay** to the viewport showing the live camera **Position** and **Target** (focal point), so the current view can be reproduced elsewhere — the driving use case being `splat-transform`'s render options (`--camera <x,y,z>` / `--look-at <x,y,z>`).

- **Hidden by default** — zero viewport change unless enabled.
- Toggled two ways, following the Show Grid / `G` pattern: the **`I` key** and a **Show Camera Info** toggle in View Options (tooltip shows the shortcut; listed in the Shortcuts popup under Show).
- **Click a row to copy** the full-precision value as comma-separated `x,y,z` (no spaces), pasteable straight into splat-transform args. Shows the existing "Copied!" feedback for 1s, guarded so live updates don't overwrite it.
- Bottom-left placement, baseline-aligned with the SUPERSPLAT version label; the cursor label is hidden while the overlay is on since the Target row shows the focal point live (same value, always current).
- Setting round-trips through `.ssproj` documents and supports `?show.camera-info=true` as a URL parameter (mirrors `showCameraPoses`).
- Localized in all 9 languages.

### Implementation notes

- New `CameraInfoOverlay` component updates from the `prerender` event, reading `camera.getPose`; DOM writes only happen when the formatted values change.
- `camera.showInfo` is added to the force-render event list — the app only renders on demand (`autoRender = false`), so without it, toggling the overlay on in a static scene would show empty text until the camera next moved.
- State follows the existing `camera.showPoses` triple (`camera.showInfo` / `setShowInfo` / `toggleShowInfo`) with the boolean owned in `editor.ts` and the default in `scene-config.ts`.
- Bare `i` doesn't conflict: `Ctrl+I` (invert selection) requires Ctrl, and the shortcut handler ignores keys when focus is in a text field.

### Testing

- `npm run lint` and `npm run build` clean; all locale JSONs validated.
- Verified in the browser: default off; `I` and the View Options toggle stay in sync both directions; values update live while navigating and reflect `camera.setPose`; click-to-copy feedback restores to the latest value; `Ctrl+I` still inverts selection; typing `i` in a text input doesn't toggle; overlay rides up when the data/timeline panels open; `?show.camera-info=true` enables it at startup with the panel toggle ticked.


slimbuck:
Awesome! Could we also show the positions of the 2 point measure?

willeastcott:
> Awesome! Could we also show the positions of the 2 point measure?

LOL. The second you commented, I merged! 😢

Issue #804, opened 2026-02-12
Issue collection - Mostly Timeline related


simonbethke:
Timeline Improvements
- Sometimes you want to have the same position that you have in the current frame also in another frame. But if you select the other frame the camera jumps there. I could imagine, that by clicking onto a frame-marker while holding the CTRL key, it would set the current camera position to the clicked frame.
- For some applications (especially rendering video) it is a pain, that the end of the animation is connected to the beginning. A first and important feature would be have a toggle switch to set if the end should be smoothed to the beginning or not.
- Instead of the evenly spaced frame-numbers, I would like to see decimal-based framenumbers and show the intermediate steps with ticks. For very few frames you'd have 1 2 3 4 5, then 2 4 6 8 10, then 5 10 15 20 25, then 10 20 30 40 50, then 20 40 60 80 100, ...
- When dragging a keyframe, the camera with the playhead that is in the middle between the dragged and another keyframe should move according to the smoothing between the dragged and the static frame.
- The camera-path needs to be visualized on the screen. Keyframes should become draggable similar to measurement-points
- The FPS should be changed to "Playback-Speed" with a float value between 0.01 and 100. The current thing is paradox: if you show 10 frames at 24 fps, every single frame becomes 2.4 frames... what you wanted to say is that the timeline doesnt show frames, but steps and the selection that you make is not frames per second, but steps per second. And once you know how many steps you have to complete per second the FPS is defined either by what you set when rendering the video or by what your computer can produce in realtime. **Other Idea:** Other professional software does it differently: They dont have frames on their timeline but "time" :D So you define the frames on a stream of seconds or milliseconds. Instead of the FPS-selection, you would then rather implement a function that materializes a speed up or slow-down as in moving all keyframes.
- Something that is currently not possible is to animate accurate circular orbits - _A very advanced thing would be if each frame knows if its created in fly or in orbit mode: Currently all frames are seen as fly-mode with the camera-center as point on the path. This makes it impossible to create circular orbits as animation. I could imagine, that there is fly and orbit-frames._
- FOV needs to be animatable aswell
- When defining keyframe 100 and then shorten the timeline length to 80, the frame 100 silently exists but is not used. I am not sure if this is really intuitive

Bugs
- When defining only one frame, it is not possible to jump back to that position
- When in fly-mode and switching to orthogonal projection, the mouse-wheel zoom doesnt work

Other Improvements
[...]

PR #942, opened 2026-07-04
Make FOV auto-dolly optional (default off) so FOV changes act as a lens zoom

willeastcott:
## Summary

From #804: *"When in fly-mode, changing the FOV should not dolly the camera forth and back."* Reworked per review feedback to make the auto-dolly behavior **optional in both modes, default off**.

Camera position is derived per frame as `distance * sceneRadius / fovFactor`, so changing FOV moved the camera along the view direction (a framing-preserving dolly). Now:

- **Default:** changing FOV acts as a lens zoom — the camera stays fixed in world space, in both orbit and fly modes. Implemented by rescaling the distance tween (`value`/`source`/`target`, clamped to zoom limits) in `setCameraFov`.
- **"FOV Auto Dolly" toggle** (View Options, under the FOV slider): restores the previous framing-preserving dolly, in both modes. Persisted in the document with the other view settings; localized in all 9 locales.

Animation playback and document load set FOV through other paths and re-derive distance themselves, so they're unaffected.

Note this changes the default orbit-mode behavior (previously always dollied to preserve framing).

## Test Plan

- [ ] Default (toggle off): sweep the FOV slider in orbit and in fly → rendered camera position does not move; image zooms like a lens
- [ ] Toggle "FOV Auto Dolly" on: FOV slider keeps the subject framed (camera dollies) in both modes
- [ ] Save/reload a document → toggle state persists
- [ ] Play a keyframed animation with varying FOV → playback unchanged


willeastcott:
@slimbuck Not sure if this behavior is 'as designed' so I won't merge before you have had a chance to comment on it.

slimbuck (commented):
I think we probably want the "auto dolly" behaviour to be optional in both modes (and probably default off).

slimbuck (approved):

PR #947, opened 2026-07-05
fix: clear error when WebGPU adapter is unavailable during Viewer/SOG export

### LINKED ISSUE #934: Cannot read properties of null (reading 'features') while saving file
Hi, when i try export 'Viewer APP->HTML or ZIP file', I met this error, my system is ubuntu20,04 and i use chrome browser, the supersplat's version is v2.28.1;

Image

I added two keyframe in timeline, and i want to get the interpolation camera poses, so that i can render rgb and depth in my own gaussian project use this greate smooth pose, colud you please tell me how can i get this poses?
Is this the final interpolation camera poses?
https://github.com/playcanvas/supersplat/blob/d54ecf301987cd3e36dfe6ca1bd0083aa6e6dffe/src/camera-poses.ts#L233-L239

mvanhorn:
## Summary

Exporting a scene to the Viewer App or SOG on a machine with no compatible WebGPU adapter now shows a clear, actionable error instead of a confusing "Cannot read properties of null (reading 'features')" crash shown under an "Error Loading File" header.

## Why this matters

As reported in #934, the Viewer/SOG export path builds a WebGPU device via `createGpuDevice` in `src/splat-serialize.ts`, which only guards against a missing `navigator.gpu`, not against `navigator.gpu.requestAdapter()` returning `null`. On systems with no compatible WebGPU adapter (common on older Linux/Chrome setups), the engine then reads `adapter.features` on a null adapter, producing the `reading 'features'` crash. This change adds an adapter pre-flight that throws a clear error before that null access is reached.

Separately, the export catch block in `src/file-handler.ts` labelled a save/export failure with the load-error header (`popup.error-loading`, "Error Loading File"), which is exactly what the maintainer found confusing in the issue thread. It now uses a new `popup.error-saving` ("Error Saving File") key. The genuine load-error uses elsewhere are left untouched, and the eight non-English locale files inherit the new key automatically through the existing `fallbackLng: 'en'` config.

## Testing

The repo has no unit-test framework, so verification is via the repo's own gates plus static reasoning about the null-adapter path:

- `npm run lint` passes clean.
- `npm run build` (type-checked rollup build) completes successfully.
- Happy path: with a working adapter, the added pre-flight `requestAdapter()` returns non-null and control proceeds unchanged into `graphicsDevice.createDevice()`.
- Bug path: with `navigator.gpu` present but `requestAdapter()` resolving `null`, `createGpuDevice` now throws "WebGPU is not available: no compatible GPU adapter was found" instead of the engine crash.
- The existing `if (!navigator.gpu)` guard is unchanged, so the no-WebGPU-at-all path still reports "WebGPU is not available in this browser".

Fixes #934

### DIFF
--- a/src/file-handler.ts
+++ b/src/file-handler.ts
@@ -587,7 +587,7 @@ const initFileHandler = (scene: Scene, events: Events, dropTarget: HTMLElement) 
         } catch (error) {
             await events.invoke('showPopup', {
                 type: 'error',
-                header: i18n.t('popup.error-loading'),
+                header: i18n.t('popup.error-saving'),
                 message: `${error.message ?? error} while saving file`
             });
         } finally {
--- a/src/splat-serialize.ts
+++ b/src/splat-serialize.ts
@@ -1129,6 +1129,14 @@ const createGpuDevice = async (): Promise<WebgpuGraphicsDevice> => {
         throw new Error('WebGPU is not available in this browser');
     }
 
+    // Ensure a compatible WebGPU adapter exists before creating the device.
+    // Without this pre-flight, a null adapter causes the engine to read
+    // adapter.features on null, crashing with a confusing error during export.
+    const adapter = await navigator.gpu.requestAdapter();
+    if (!adapter) {
+        throw new Error('WebGPU is not available: no compatible GPU adapter was found. This export format requires WebGPU.');
+    }
+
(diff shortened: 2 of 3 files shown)


slimbuck:
Thanks for this PR too! Closing in favour of #956

PR #1012, opened 2026-08-25
Fix camera pose import memory leak

LSQzzx:
## Summary

- Batch-load JSON and COLMAP camera poses instead of rebuilding the animation track for every pose.
- Preserve the default camera FOV when imported poses do not provide one.
- Correctly unregister tooltip listeners using the DOM element supplied by the destroy event.
- Release destroyed timeline key wrappers from the tooltip registry.

## Root cause

Camera imports emitted one `track.keyAdded` event per pose. Each event rebuilt every timeline key, resulting in O(n²) DOM allocation.

Destroyed PCUI elements clear `target.dom` before emitting `destroy`. Tooltip cleanup then attempted to access the cleared DOM, preventing entries from being removed from its internal map and retaining millions of destroyed wrappers.

## Testing

- `npm run lint`
- `npm run build`
- Verified that importing 3,000 poses emits one `track.keysLoaded` event and no per-pose `track.keyAdded` events.

### DIFF
--- a/src/camera-poses.ts
+++ b/src/camera-poses.ts
@@ -204,8 +204,10 @@ class CameraAnimTrack implements AnimTrack {
      * Load poses from serialized data.
      */
     loadPoses(posesData: Pose[]): void {
+        const defaultFov = this.events.invoke('camera.fov') ?? 60;
         this.poses.length = 0;
         posesData.forEach((pose) => {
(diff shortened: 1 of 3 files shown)


slimbuck:
Thanks for this, I've applied this fix to latest version of editor in #1019.

PR #1029, opened 2026-09-06
Add Min Splat Size control to Settings panel

jacomof:
The projected-splat-size cull threshold was hardcoded to 2px in editor.ts with no UI to change it. Scenes with small or thin gaussians lose visible points once the camera is far enough that splats project under this threshold, with no way to disable or tune it.

Adds a "Min Splat Size" slider (0-20px) to Preferences > Rendering, next to SH Bands, wired the same way as the existing sliders there.

### DIFF
--- a/src/preferences.ts
+++ b/src/preferences.ts
@@ -93,6 +93,7 @@ const registerPreferences = (events: Events, config: SceneConfig, urlArgs: any) 
         { key: 'camera.fovDolly', setCommand: 'camera.setFovDolly', getDefault: () => false, validate: isBool, group: 'preferences' },
         { key: 'camera.fov', setCommand: 'camera.setFov', urlPath: 'camera.fov', getDefault: () => config.camera.fov, validate: isNumber(10, 120), group: 'preferences' },
         { key: 'view.bands', setCommand: 'view.setBands', urlPath: 'show.shBands', getDefault: () => config.show.shBands, validate: v => typeof v === 'number' && Number.isInteger(v) && v >= 0 && v <= 3, group: 'preferences' },
+        { key: 'view.minPixelSize', setCommand: 'view.setMinPixelSize', getDefault: () => 2, validate: isNumber(0, 20), group: 'preferences' },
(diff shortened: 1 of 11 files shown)


slimbuck:
Hello, could you describe why you need this please? Thanks!

willeastcott:
@jacomof One thing that often helps a lot is if you attach a screen cap of the update in the PR description to help people understand the change.

jacomof:
Hello @slimbuck and @willeastcott,

Thank you for getting back to me and I'm sorry for the not very descriptive initial PR. I developed a method for 3D underwater scene generation and I was trying to visualize its results using supersplat. While it worked well, I encountered issues when trying to visualize entire scenes, as there is a fixed threshold (minPixelSize, src/editor.ts:1259-1268) on the pixel size of a single guassian, below which it becomes invisible. For instance, see the image below:

scene-w-holes


minPixelSize was already exposed through 'view.minPixelSize' getter/setter events, I just hooked it to a slider in the UI.  I think this may help other people with small optimized gaussians. For instance, reducing the threshold to 1 makes the previous view fully visible:


scene-wout-holes

I did not find any existing setting or tweak to make the scene completely visible, but please let me know if there is an standard practice to solve this issue.

slimbuck:
Hi @jacomof,

Thanks for the explanation. Your use-case makes a lot of sense. It's strange just how different synthetic gaussian splat data behaves compared to trained ones.

I think we must add this option to the editor, but I wonder about superspl.at and the viewer.

jacomof:
Good question @slimbuck. Just to clarify, this PR only touches the editor's own custom splat-drawing layer (projected-splat-renderer.ts / projected-splat-projector-shader.ts). 

The viewer (and I'd assume superspl.at's viewer page too) doesn't seem to use that layer at all, from what I can gather it's built on the PlayCanvas engine's native GSplatMaterial instead, which has its own hardcoded minPixelSize = 2 living in the playcanvas/engine repo. So it's a genuinely separate code path.

I don't have a strong opinion on whether that should also be made configurable in the viewer as I have not used it much, that'd be a separate change in playcanvas/engine, and I haven't looked at what else depends on that value there or what the tradeoffs would be. Happy to try implementing it as a follow-up if you think it's worth it.

PR #1030, opened 2026-09-06
Add object transform keyframe track, independent of the camera

jacomof:
The timeline could only ever keyframe the camera. track-manager.ts's getActiveTrack() was hardcoded to the camera track, with a comment noting selection-based switching wasn't implemented yet. There was no way to animate a splat's own position/rotation/scale (e.g. a turntable where the model spins in place rather than the camera orbiting it).

Adds a Camera/Object mode toggle to the timeline panel. In Object mode, Add Key / Remove Key / drag-to-move-or-copy operate on a new SplatAnimTrack for the selected splat instead of the camera track. SplatAnimTrack keyframes the splat's local position/rotation/scale and interpolates with plain lerp/slerp between the two neighbouring keys (not a fitted spline), which stays monotonic and overshoot-free regardless of key spacing. This matters for a full-turn spin, where a spline fitted through few keys around a circle can wobble. Both the camera track and any per-splat tracks evaluate every frame regardless of the active mode, so a camera move and an object spin can run together.

Known limitations, flagged for discussion:
- Object keyframes are not yet saved/loaded with the document (only in-session).
- The rotate gizmo's pivot defaults to the splat's local origin, not its visual center (existing behavior, Splat.localFrameOrigin defaults to zero). For a clean in-place spin, set a local frame via the Orient Tool first. To be determined if the Object mode should set the origin automatically.


slimbuck:
Hello, can you explain why you need this or how you use it? Thanks

jacomof:
Hello @slimbuck,

I am sorry for the delay and the not very descriptive PR. I developed a method/model to generate 3D underwater terrain scenes and I wanted to do a 360 degrees visualization of the environment. However, I did not find an easy way in the UI to  record key frames that tracked the object rotating itself, rather than the camera. When you rotate an object and add a key frame, the change is not registered in the video generated by supersplat (see the first video below). Due to camera smoothing and the size of the scene, it was hard to recreate the same rotation effect by rotating the camera around the object. For this reason, I think it would be very helpful to track the movement of the objects as well as the camera when capturing key frames, which will give more flexibility and ease-of-use to the users when capturing object-centric videos. In the PR, I have modifed the key frame recording tab with an object mode, that records movements of the object. I decided not to combine the object and camera recording modes in a single tab to maximize compatibility for previous users.

Video of object rotation not being recorded:
  

https://github.com/user-attachments/assets/6eb3aebf-5fc2-4bf7-8874-9fa4f65d9460


Video of the PR change that allows rotation:


https://github.com/user-attachments/assets/3c43f4e0-b187-4eb6-b617-bbabcf7557d0

PR #1036, opened 2026-09-08
Add Spherical Beta view-dependent color with diffuse/specular separation and PLY export

RongLiu-Leo:
### Highlight

Add Spherical Beta view-dependent color with diffuse/specular separation and PLY export

---

Adds support for Spherical Beta (SB), the view-dependent color model from
[Deformable Beta Splatting](https://rongliu-leo.github.io/beta-splatting/)
(Liu et al., SIGGRAPH 2025). SB replaces the higher SH bands with a few explicit
specular lobes, matching or beating SH quality with far fewer per-splat values —
2 lobes (12 floats) against SH bands 1–3 (45 floats).

Trainers write the lobes to PLY as `sb_params_*` columns; this makes SuperSplat
read, render, edit and re-export them. Files without those columns are
unaffected — SB is detected per-file and every shader path sits behind
`#if SB_LOBES > 0`.

### Rendering

`sb-utils.ts` packs the lobes into an RGBA32F atlas (2 texels per lobe, indexed
by `splat.index`); `evalSB` in the splat shader adds them to the SH color:

```
C += sum_l pow(max(dot(dir, lobeDir_l), 0), 4 * exp(beta_l)) * softplus(rgb_l)
```

SH evaluation is untouched and the two share one view direction, so a file may
carry either model or both.

### UI

- **SB lobes** slider beside the existing SH bands slider.
- **Color mode** selector (Full / Diffuse / Specular), reproducing the paper's
  light decomposition. Diffuse compiles out both view-dependent terms; Specular
  substitutes the background for the band-0 term so splats with no
  view-dependent energy blend away.

### Export

`splat-serialize.ts` writes `sb_params_*` back out, handling the parts that
aren't a passthrough:

- **Mixed lobe counts on merge** — columns are channel-major
  (`sb_params_{channel * lobes + lobe}`), so fewer-lobe sources need index
  remapping, not copying.
- **Absent lobes** — 0 isn't neutral (`softplus(0)` ≈ 0.1 would add a visible
  specular term), so they're padded with a raw value that activates to zero.
- **Baked transforms** — a lobe is a direction, so rotation applies to it
  directly. The SB analogue of the existing SH rotation, and much cheaper.
- `maxSBLobes` export setting mirrors `maxSHBands`.

### Compatibility

Lobes are additive properties; the standard `f_dc_*` / `f_rest_*` schema is
untouched, so viewers without SB still load the file and render the diffuse
base. A `sb_params_*` count that isn't a whole number of lobes, or exceeds the
supported count, is rejected with a warning rather than partially read — a
partial read would misinterpret one channel as another.

### Notes for review

- `SB_MAX_LOBES` is 2, covering the paper's configuration. It bounds the shader
  loop and atlas size, not the model; raising it is a one-line change.
- The `> 0` in `#if SB_LOBES > 0` is deliberate: PlayCanvas's preprocessor treats
  bare `#if NAME` as `#ifdef NAME`, so a 0-valued define would still be taken.
- Locale strings added for all nine supported languages.

### Testing

---

## Before opening

- Branch is **26 commits behind** upstream `main` — rebase first.
- Drop `.github/workflows/deploy-pages.yml` (fork-only GitHub Pages deploy).
  Takes the PR from 4 commits to 3 cleanly scoped ones.
- Cross-link with the gsplat SB PR once both are open.


RongLiu-Leo:
**Live build of this branch:** https://rongliu-leo.github.io/supersplat/

Sample PLYs to try it with — drag either in:

- [SH0_SB2.ply](https://drive.google.com/file/d/100z7Dw6OdpDDnW5m2RCcuFPe_TbIZekh/view?usp=drive_link) — carries Spherical Beta lobes
- [SH3_SB0.ply](https://drive.google.com/file/d/1QHkVx2vRs8zslxCzpCmj25gLv6WsLBtA/view?usp=drive_link) — plain SH, to confirm non-SB files are unaffected

The producer side is up as nerfstudio-project/gsplat#1056, which adds Spherical
Beta training to gsplat and writes the `sb_params_*` columns this PR reads. For
context on why the format is worth supporting: on MipNeRF-360, 2 lobes beat SH
degree 3 on PSNR at ~45% of the PLY size and roughly half the peak VRAM.

### Render Comparisons: SB has a better view dependency separation than SH

SH (top) vs SB (bottom)
#### Full

Screenshot 2026-09-08 173901
Screenshot 2026-09-08 173917

#### Diffuse

Screenshot 2026-09-08 173933
Screenshot 2026-09-08 173946

#### Specular

Screenshot 2026-09-08 173958
Screenshot 2026-09-08 174010

